PR #93, opened 2024-10-04
Implement Http Client Logic

### LINKED ISSUE #89: Implement HTTP client logic
## Description

As part of the Command Manager plugin development and in continuation of #73, we are going to implement an HTTP client service to communicate with the Wazuh Server Management API.

This HTTP client service will implement and expose a method to perform async POST requests, taking commands as payload.

```java
httpService.post(order);
```

## Functional requirements

- The HTTP Client service implements a method to perform async POST requests to the Wazuh Server Management API.
- The HTTP Client service POST method takes a command as in its index model as parameter. This will be sent as payload.
- The HTTP Client service returns information about the request's completion or failure.
- The HTTP Client service uses appropriate logging.
- The HTTP Client service handles connections errors.

- The HTTP client service will use Apache HTTP Client under the hood.
- The Wazuh Server Management API details are to be read from the configuration.

f-galland:
### Description
This PR adds Http client functionality to the command manager plugin, meant to query the Wazuh Server's command API.

### Issues Resolved
Resolves #89


f-galland:
I've attempted to migrate towards `HttpClient` but I've found that it depends on [SLF4J](http://www.slf4j.org/) which seems not to be compatible with `org.apache.logging.log4j.Logger` (which is the logging library we've been using so far).

We might have to move all the plugins to `slf4j` for consistency's sake.

f-galland:
Actually, just adding `slf4j` to the `build.gradle` following the example in `HttpClient`'s [pom](https://github.com/apache/httpcomponents-client/blob/master/pom.xml) gets rid of the issue.

AlexRuiz7 (approved):

Issue #171, opened 2024-12-06
Add Management API CA file field to `opensearch.yml`


f-galland:
## Description

In order for the command-manager plugin to connect to the Wazuh Server's management API using self-signed certificates, a `CAcert` needs to be loaded into the plugin's trust chain.
We need to expose this as a setting in the configuration file.

f-galland:
A sample of a class reading a config file from a plugin:
https://github.com/opensearch-project/reporting/blob/2.18.0.0/src/main/kotlin/org/opensearch/reportsscheduler/settings/PluginSettings.kt

A sample class loading a custom trust chain:
https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientCustomSSL.java

mcasas993:
Relevant discovery in the process:

### How to put configuration attributes in opensearch.yml.
The process is simple and similar to putting configurations in keystore. Depending on the type of attribute, if it is a secure configuration or not, it will automatically be saved in the keystore or in the opensearch.yml file.

Example of declaration of secure setting:
```
 public static final Setting<SecureString> M_API_AUTH_USERNAME =
            SecureSetting.secureString("m_api.auth.username", null);
```

Example of declaration of not secure setting:
```
public static final Setting<String> WAZUH_INDEXER_CA_CERT_PATH =
        Setting.simpleString("ssl.http.pemtrustedcas_filepath", Setting.Property.NodeScope);
```

### Validate configuration values.
We do not yet validate the contents of the attributes in the settings, but we will probably want to do so in the future. We can see a good example of how to create and validate settings [here](https://github.com/opensearch-project/OpenSearch/blob/e46d4bc47423a6bcf61182dbc42cb5d68e5cb61b/server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java#L391).

### Access to files with AccessController.doPrivileged
To use the AccesController we have to configure a policy file inside plugins/command-manager/src/main/plugin-metadata/

Inside this file, if we want to manage the access to a file we cannot put a “*”, we have to put a concrete directory or an absolute path to a file.

Example.
`    grant {
permission java.io.FilePermission "/home/user/Documents/wazuh-certificates/root-ca.pem", "read";}
`

PR #179, opened 2024-12-10
Connect to the Wazuh Server's management API using self-signed certificates

mcasas993:
### Description
In order for the command-manager plugin to connect to the Wazuh Server's management API using self-signed certificates, a CAcert needs to be loaded into the plugin's trust chain.
We are going to expose this as a setting in the configuration file.

### Issues Resolved
[[171](https://github.com/wazuh/wazuh-indexer-plugins/issues/171)]


mcasas993:
At this moment, the settings figure in opensearch.yml 

image

PR #235, opened 2025-01-20
Log settings on plugin's start

### LINKED ISSUE #225: [Feature request] Add logging when custom settings are used
**Is your feature request related to a problem? Please describe.**
During the testing of #180 it was hard noticing if custom settings were in use, specially when debugging some of the other issues detected.

**Describe the solution you'd like**
It would be great to have some logging about the use of custom values for the CM's settings. That way, we can easily see which settings are in use at a given point in time. Useful for debugging (for development and production). 

**Additional context**
Probably these logs would be of `debug` level.

AlexRuiz7:
### Description
This PR logs the plugin's settings on start up.

```log
[c.w.c.s.PluginSettings   ] [integTest-0] Settings loaded: {timeout=2, jobSchedule=1, jobPageSize=100, jobKeepAlive=30}
```

### Issues Resolved
Closes #225

### DIFF
--- a/plugins/command-manager/src/main/java/com/wazuh/commandmanager/settings/PluginSettings.java
+++ b/plugins/command-manager/src/main/java/com/wazuh/commandmanager/settings/PluginSettings.java
@@ -82,6 +82,7 @@ public class PluginSettings {
         this.jobSchedule = JOB_SCHEDULE.get(settings);
         this.jobPageSize = JOB_PAGE_SIZE.get(settings);
         this.jobKeepAlive = JOB_KEEP_ALIVE.get(settings);
+        log.error("Settings loaded: {}", this.toString());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


plugins/command-manager/src/main/java/com/wazuh/commandmanager/settings/PluginSettings.java
```diff
@@ -82,6 +82,7 @@ private PluginSettings(@NonNull final Settings settings) {
         this.jobSchedule = JOB_SCHEDULE.get(settings);
         this.jobPageSize = JOB_PAGE_SIZE.get(settings);
         this.jobKeepAlive = JOB_KEEP_ALIVE.get(settings);
+        log.error("Settings loaded: {}", this.toString());
```
QU3B1M, line 85: May be better if the log level is `info` or `debug`.

QU3B1M: Looking at the [issue](https://github.com/wazuh/wazuh-indexer-plugins/issues/225), maybe `debug` is the correct level for this log

AlexRuiz7: Fixed in 66074be5f391541041a592ba2007a6ed09f39abd

QU3B1M (approved):

PR #250, opened 2025-01-24
Refactor search with updateByQuery

### LINKED ISSUE #249: New commands maintenance design
### Description

In order to adapt to the [command manager's latest design](https://github.com/wazuh/wazuh-indexer/issues/349) requirements, we need to rewrite the `SearchThread` class, since most of it is tailored to the outdated data flow design.

We will use an [updateByQuery](https://opensearch.org/docs/latest/api-reference/document-apis/update-by-query/) style API call to simultaneously retrieve pending commands and update their status to `failure` whenever these happen to have expired.

f-galland:
### Description
This PR replaces the `SearchThread` class with a new one that uses `updateByQuery` to conform to the latest command-manager design requirements.

### Issues
Closes #249


f-galland:
It looks like this works ok even for larger numbers of commands. I tested this with the following script:

```bash
#!/bin/bash

command='{"source":"Engine","user":"user53","target":{"id":"target4","type":"agent"},"action":{"name":"restart","args":{"arg1":"/path/to/executable/arg6"},"version":"v4"},"timeout":30}'
commands='{"commands": ['

for i in {1..744}
do
  commands+="$command"
  commands+=","
done

commands=${commands::-1}
commands+=']}'

echo $commands > commands.json

curl http://localhost:9200/_plugins/_command_manager/commands -H 'Content-Type: application/json' -d "$(cat commands.json)"
```

The number of commands, 744, was chosen as it was the maximum curl would allow me to send at once. I fired this repeatedly and then waited for the timeouts.

I then queried the engine for commands with `failure` status:

```bash
#!/bin/bash


echo -ne 'Failed:\n\n'
curl http://localhost:9200/.commands/_count -H 'Content-Type: application/json' -d '
{
  "query": { 
    "bool": { 
      "filter": [ 
        { "term":  { "command.status": "failure" }}
      ]
    }
  }
}
'
echo -ne '\n\nPending:\n\n'
curl http://localhost:9200/.commands/_count -H 'Content-Type: application/json' -d '
{
  "query": { 
    "bool": { 
      "filter": [ 
        { "term":  { "command.status": "pending" }}
      ]
    }
  }
}
'
```

This is what I got:
```shellsession
Failed:

{"count":3720,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0}}

Pending:

{"count":0,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0}}
```

AlexRuiz7 (approved):

Issue #248, opened 2025-01-24
Dedicated `command.action.args` models


AlexRuiz7:
## Description

We already have a [solid definition](https://github.com/wazuh/wazuh-agent/blob/master/docs/ref/commands.md) for the first set of commands that will be accepted by Wazuh Agents.

Currently, the Command Manager plugins does not validate per command type, and does a generic validation instead. For the commands of type `set-group` and `fetch-config`, the Command Manager needs to do a proper validation of the fields so they match the given definition model, specially these under the `command.action.args` object.

The `set-group` command is used to update the group of agents that the agent belongs too. This command accepts an **array of strings** within `action.args`, comprising the complete list of groups for this agent. 

The `fetch-config` command is used to order an agent to download and apply the centralized configuration for the groups it belongs too. This command accepts **no arguments**.

## Functional requirements

- In case of a mismatch on a `set-group` command, the request is rejected.
- In case of a mismatch on a `fetch-config` command, mismatching values are ignored, and the request is accepted.
- For any other command type, the request is accepted under the generic parser.

## Implementation restrictions

- Unit tests required as PR acceptance criteria.

## Plan

- [x] Create a dedicated parser (or model) for `set-group` commands.
- [x] Create a dedicated parser (or model) for `fetch-config` commands.

PR #253, opened 2025-01-29
Add hard validation to set-group and fetch-config commands

mcasas993:
### Description
Create a dedicated model for set -group and fetch-config commands.

The set-group command accepts an array of strings within action.args, comprising the complete list of groups for this agent. In case of a mismatch on this command, the request will be rejected.

The fetch-config command accepts no arguments. In case of a mismatch on this command, mismatching values are ignored, and the request is accepted.

### Issues Resolved
[248](https://github.com/wazuh/wazuh-indexer-plugins/issues/248)


AlexRuiz7 (changes requested):
I found a bug where the action is incorrectly parsed using the generic parser. This happens on commands where `action.args` is specified before than `action.name`, has `action.name` has not been parsed yet and the current logic in unable to determine the action type.


```java
        String name = "";
        Args args = new Args();
        String version = "";

        while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
            String fieldName = parser.currentName();
            parser.nextToken();
            switch (fieldName) {
                case NAME:
                    name = parser.text();
                    break;
                case Args.ARGS:
                    switch (name) {
                        case "set-group":
                            log.info("Parsing arguments for [set-group] command");
                            args = SetGroupCommand.parse(parser);
                            break;
                        case "fetch-config":
                            log.info("Parsing arguments for [fetch-config] command");
                            args = FetchConfigCommand.parse(parser);
                            break;
                        default:
                            log.info("name: {}", name);
                            log.info("Parsing arguments for [generic] command");
                            args = Args.parse(parser);
                            break;
                    }
                    break;
```

🟢 Parsed as `set-group`, as `action.name` is read first.
```json
{
  "commands": [
    {
      "action": {
        "name": "set-group",
        "args": {
          "groups": [
            "group_1",
            "group_2"
          ]
        },
        "version": "5.0.0"
      },
      "source": "Users/Services",
      "user": "Management API",
      "timeout": 100,
      "target": {
        "id": "d5b250c4-dfa1-4d94-827f-9f99210dbe6c",
        "type": "agent"
      }
    }
  ]
}
```

🔴 Parsed as `generic`, as `action.args` is read first.
```json
{
  "commands": [
    {
      "action": {
        "args": {
          "groups": [
            "group_1",
            "group_2"
          ]
        },
        "name": "set-group",
        "version": "5.0.0"
      },
      "source": "Users/Services",
      "user": "Management API",
      "timeout": 100,
      "target": {
        "id": "d5b250c4-dfa1-4d94-827f-9f99210dbe6c",
        "type": "agent"
      }
    }
  ]
}
```

AlexRuiz7 (approved):

AlexRuiz7:
I'm approving this PR, and we'll figure out how to redirect the request to the proper parser unambiguously on a dedicated issue.

PR #294, opened 2025-02-27
Changes empty values for null (indexer-plugins)

abbonno:
### Description
In order to avoid indexing void data, empty arrays and strings such as `""` and `[]` are modified by `null`.

### Issues Resolved
[Management of null values on indices #593](https://github.com/wazuh/wazuh-indexer/issues/593)


AlexRuiz7:
The goal of the issue is not to modify our plugins' code, but to demand other components to clean their data before indexing, avoiding sending empty values.

PR #302, opened 2025-03-06
[ECS Generator] Update index templates

wazuh-devel-xdrsiem-indexer:
This PR updates the ECS templates for the following modules:
- states-vulnerabilities


AlexRuiz7:
No changelog entry needed as the change was never published (we are reverting the change before the release), and there was no changelog entry for its addition (so no removal required).

f-galland (approved):

PR #471, opened 2025-05-30
Add ISM rollover policy for stateless indices

### LINKED ISSUE #466: Create ISM rollover policy for stateless indices

f-galland:
### Description
This PR loads an ISM policy from the setup plugin. This automatically creates aliases for indices where its required and instructs the ISM plugin to automatically rollover the target index to a new one as the conditions to rotate are met.

### Issues Resolved
Closes #466

### DIFF
--- a/plugins/setup/src/test/java/com/wazuh/setup/index/IndexStateManagementTests.java
+++ b/plugins/setup/src/test/java/com/wazuh/setup/index/IndexStateManagementTests.java
@@ -41,7 +41,6 @@ public class IndexStateManagementTests extends OpenSearchTestCase {
 
     private IndexStateManagement ismIndex;
     private Client client;
-    private AdminClient adminClient;
     private IndicesAdminClient indicesAdminClient;
     private IndexUtils indexUtils;
 
@@ -50,7 +49,7 @@ public class IndexStateManagementTests extends OpenSearchTestCase {
         super.setUp();
 
         client = mock(Client.class);
-        adminClient = mock(AdminClient.class);
+        AdminClient adminClient = mock(AdminClient.class);
         indicesAdminClient = mock(IndicesAdminClient.class);
         indexUtils = mock(IndexUtils.class);
 
@@ -65,6 +64,8 @@ public class IndexStateManagementTests extends OpenSearchTestCase {
     /**
      * Verifies that {@link IndexStateManagement#initialize()} creates the index and indexes ISM
      * policies when the index does not already exist.
+     *
+     * @throws IOException if an error occurs while reading the policy file
      */
     public void testInitialize_CreatesIndexAndPolicies() throws IOException {
         Map<String, Object> template = new HashMap<>();
@@ -99,7 +100,7 @@ public class IndexStateManagementTests extends OpenSearchTestCase {
      * Verifies that if the index already exists, {@link IndexStateManagement#initialize()} skips
      * index creation.
      */
-    public void testIndexAlreadyExists_SkipsCreation() throws IOException {
+    public void testIndexAlreadyExists_SkipsCreation() {
         doReturn(true).when(ismIndex).indexExists(".opendistro-ism-config");
 
         doReturn(mock(ActionFuture.class)).when(client).index(any(IndexRequest.class));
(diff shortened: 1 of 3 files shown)


f-galland:
I've been able to reproduce the issue after generating a package locally and installing it on a VM.

The error itself is thrown from the `Bootstrap.java` class:
* https://github.com/opensearch-project/OpenSearch/blob/dc4efa821904cc2d7ea7ef61c0f577d3fc0d8be9/server/src/main/java/org/opensearch/bootstrap/Bootstrap.java#L455

But I haven't been able to pin the source of the exception. It does look like the ISM plugin is trying to write its own index after a restart, but I haven't been able to double check whether that's the case.

f-galland:
The issue was caused by using a `RoutingTable` object obtained at cluster startup. This object is not updated, but a new one is created each time the table changes.
The correct way to get an updated routing table is to call `ClusterService.state().getRoutingTable()`.
This fixed the issue.
Thanks @AlexRuiz7

Jorgesnchz (changes requested):
I have reviewed the code and everything seems to be good. The only issue I found is that the last functional requirements of the Issue are not being explicitly fulfilled.

The value of Minimum primary shard size is supposed to be 25GB, but since it is not declared, I believe it defaults to 10GB [(Reference)](https://www.elastic.co/docs/deploy-manage/production-guidance/optimize-performance/size-shards) 

The same applies to the minimum age of the index, which is not specified as 7 days, and to the Minimum document count, which does not have the value of 200M specified.

plugins/setup/src/main/java/com/wazuh/setup/utils/IndexUtils.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Cast map's element to a String, Object map.
+     *
+     * <p>Used to retrieve the settings and mappings from the index templates, which are a JSON object
+     * themselves.
+     *
+     * @param map the index template as a map.
+     * @param key the element's key to retrieve and cast.
+     * @return a String, Object map
+     */
```
Jorgesnchz, line 75: Missing `@SuppressWarnings("unchecked")`, since we use in other files I believe we should use it in all of them

Jorgesnchz (changes requested):
I tested it in a production environment and in a real environment everything is working good

There are a few improvements that can be done, but everything is working fine

Jorgesnchz:
Also, in order to merge this PR there should be some documentation of the changes that it involves and how does it work in order to keep the code easier to maintain

plugins/setup/src/main/java/com/wazuh/setup/index/IsmIndexInitializer.java
```diff
@@ -0,0 +1,162 @@
...
+                            .id(SetupPlugin.WAZUH_ALERTS_ROLLOVER_POLICY_ID)
+                            .source(policy, MediaTypeRegistry.JSON);
+
+            client.index(indexRequest).actionGet(SetupPlugin.TIMEOUT);
+            log.info("Indexed Wazuh rollover policy into {} index", indexStrategySelector.getIndexName());
+        } catch (IOException e) {
+            log.error("Failed to load the Wazuh rollover policy from file: {}", e.getMessage());
+        } catch (ResourceAlreadyExistsException e) {
+            log.error("Policy already exists, skipping creation: {}", e.getMessage());
+        }
```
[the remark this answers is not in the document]
f-galland: I purposefully avoided performing that check here. The rationale being that even if the index or template were not created for some reason, we will still _try_ to index the policy.

plugins/setup/src/main/java/com/wazuh/setup/utils/IndexUtils.java
f-galland: This annotation avoids warnings in `gradle check`. Even if there is an unsafe cast here, its not stopping our tests from going through. I don't think its really necessary.

f-galland:
> Also, in order to merge this PR there should be some documentation of the changes that it involves and how does it work in order to keep the code easier to maintain

That objective is part of a separate issue:
* https://github.com/wazuh/wazuh-indexer-plugins/issues/489

Jorgesnchz (approved):
LGTM

Jorgesnchz:
In a real environment, it was tested to install the plugin and once It was installed, restart the service by doing a `systemctl restart wazuh-indexer.service`

And everything worked well after it

Jorgesnchz (approved):

AlexRuiz7 (approved):

PR #484, opened 2025-06-10
Reduce risk of `GITHUB_TOKEN` exposure

Jorgesnchz:
### Description
This PR adds permissions to the workflows of main branch that use `GITHUB_TOKEN` to reduce the risk of exposure of the token

### Issues Resolved
Closes https://github.com/wazuh/internal-devel-requests/issues/2507

### DIFF
--- a/.github/workflows/5_builderpackage_plugins.yml
+++ b/.github/workflows/5_builderpackage_plugins.yml
@@ -52,7 +52,6 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       actions: read
-      contents: write
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/5_builderpackage_plugins.yml
```diff
@@ -50,6 +50,9 @@ on:
 jobs:
   build:
     runs-on: ubuntu-24.04
+    permissions:
+      actions: read
+      contents: write
```
AlexRuiz7, line 55:
```suggestion
```

AlexRuiz7 (changes requested):

PR #485, opened 2025-06-10
Reduce risk of GITHUB_TOKEN exposure

Jorgesnchz:
### Description
This PR cherry-picks the commit of main branch where permissions are added to the workflows of main branch that use `GITHUB_TOKEN` to reduce the risk of exposure of the token

### Issues Resolved
Closes https://github.com/wazuh/internal-devel-requests/issues/2507


abbonno (commented):
Wouldn't this be necessary in every workflow that uses the GITHUB_TOKEN ?

AlexRuiz7 (approved):

PR #652, opened 2025-11-11
Replace time-series indices with data streams

### LINKED ISSUE #650: Data Streams
## Description

In #644, we worked on a Proof of Concept about Data Streams as a replacement of our regular time-series indices. The PoC was very successful and uncovered big improvements compared to our current build, with minimal development effort.

- **Simplification of time-series indices creation and management**. Data Streams are designed for time-series data, so it includes features to make its management easier.
- **Introducing of Composable Index Templates**. This is a requirement for Data Streams to work. While we are using legacy index templates, Composable Index Templates allow us to reduce the size of the index templates by extracting common schemas as components (generalization and inheritance).
- **Better coupling with the UI**. The UI has dedicated sections for Data Streams and Index Templates, which are currently empty. With the addition of Data Streams and v2 Index Templates, the users can experience completeness and cohesion while navigating through the Index Management sections.
- **No drawbacks.** Every actor interacting with the Indexer, such as other Wazuh Central Components, and end users, will still be able to perform any action as before. There is no functionality loss.

## Functional requirements

- **Data Streams**. Stateless / time-series indices are replaced with Data Streams. The UI properly displays them.
- **v2 Index Templates**. Legacy index templates are replaced by v2 index templates. The UI properly displays them.
- **ISM rollover policy**. Data Streams are managed by the default rollover policy, as indices were.
- **No aliases**. With the addition of data streams, the creation of aliases is no longer required.

- **Automation**. Management of index templates and indices creation must be automated.

## Plan

- [x] Update WCS tooling to copy v2 index templates to the plugins' resources folder.
- [x] Replace creation of legacy index templates with v2 index templates. 
- [x] Replace time-series indices creation with data streams.
- [x] Clean-up legacy index templates and related stuff.
- [x] Update documentation related to time-series indices.
- [x] Update or develop new integration tests.

AlexRuiz7:
### Description
This PR replaces stateless indices with data streams.
Also, replaces legacy index templates with v2 index templates.

### Issues Resolved
Closes #650

### DIFF
--- a/ecs/stateless/access-management/fields/template-settings-legacy.json
+++ b/ecs/stateless/access-management/fields/template-settings-legacy.json
@@ -1,6 +1,6 @@
 {
   "index_patterns": [
-    "wazuh-events-v5-access-management-*"
+    "wazuh-events-v5-access-management*"
   ],
   "order": 1,
   "settings": {
--- a/ecs/stateless/access-management/fields/template-settings.json
+++ b/ecs/stateless/access-management/fields/template-settings.json
@@ -1,6 +1,6 @@
 {
   "index_patterns": [
-    "wazuh-events-v5-access-management-*"
+    "wazuh-events-v5-access-management*"
   ],
   "priority": 1,
   "data_stream": {},
--- a/ecs/stateless/applications/fields/template-settings-legacy.json
+++ b/ecs/stateless/applications/fields/template-settings-legacy.json
@@ -1,6 +1,6 @@
 {
   "index_patterns": [
-    "wazuh-events-v5-applications-*"
+    "wazuh-events-v5-applications*"
   ],
   "order": 1,
   "settings": {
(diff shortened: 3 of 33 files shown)


ecs/generator/images/generator.sh
```diff
@@ -74,14 +74,28 @@ generate_mappings() {
...
+    # Stateful templates remain unchanged except for the formatting
+    jq '{
+      "index_patterns": .index_patterns,
+      "priority": .order,
+      "template": {
+        "settings": .settings,
+        "mappings": .mappings
+      }
+    }' "$in_file" >"$out_dir/generated/elasticsearch/legacy/opensearch-template.json"
+  fi
```
QU3B1M: Maybe this could be simplified to avoid duplication since the object is almost the same (only difference is `data_stream` ), it could be something like this:


```suggestion
jq --argjson is_stateless "$([[ $ecs_module =~ stateless/ ]] && echo true || echo false)" '
{
  "index_patterns": .index_patterns,
  "priority": .order,
  "template": {
    "settings": .settings,
    "mappings": .mappings
  }
} + (if $is_stateless then { "data_stream": {} } else {} end)
' "$in_file" >"$out_dir/generated/elasticsearch/legacy/opensearch-template.json"
```

PR #702, opened 2025-12-16
Resolve SAP dependency in Content Manager's build workflow

f-galland:
### Description
This PR fixes gradle and github workflows so that SAP and its commons library are cloned, built, published to a local maven repo and then pulled from there as dependencies to build the Content Manager plugin.

### Issues Resolved
Resolves https://github.com/wazuh/internal-devel-requests/issues/3754


f-galland:
## Build workflow test run:
- https://github.com/wazuh/wazuh-indexer-plugins/actions/runs/20280080579

The on-push build workflow currently fails due to the fact that when no ref commit is declared for the SAP repo, it tries to use `main`, which still doesn't include the changes. This should change once [SAP's associated PR](https://github.com/wazuh/wazuh-indexer-security-analytics/pull/17) is merged.

.github/actions/5_builderpackage_indexer_branch_resolver/resolve_branches.sh
Jorgesnchz: We could use the branch resolver action directly from the wazuh-indexer repository action, similar to how it is done with the emails action in the sap repository ([reference](https://github.com/wazuh/wazuh-indexer-security-analytics/blob/main/.github/workflows/5_codequality_email.yml))

Jorgesnchz (changes requested):
Everything is working, and it could be merged as it is, since it is complete. But I think that it could be a good idea to use the same action instead of duping it, so if we need to apply some changes the changes are only made in one repository

Jorgesnchz (approved):

PR #707, opened 2025-12-18
Fix update of policy type documents

Jorgesnchz:
### Description
This PR Includes the logic related to the policy type documents, and avoid the `ResourceNotFoundException` from happening unless it is a real Exception

### Issues Resolved
Closes IDR3767


Jorgesnchz:
## Validation

To test it, load the plugin in a `wazuh-indexer` environment and check the `/var/log/wazuh-indexer/wazuh-cluster.log`


Log is not appearing:
``` console
[root@centos8 vagrant]# cat /var/log/wazuh-indexer/wazuh-cluster.log | grep ResourceNotFoundException
[root@centos8 vagrant]# 

```

PR #732, opened 2026-01-14
Update CTI Contexts and Consumers to v0.0.2

Jorgesnchz:
### Description
This PR updates the context and consumers to the new version

### Issues Resolved
IDR3916


Jorgesnchz:
I believe that a changelog entry is not needed, since it is an internal process that might be repeated for each new context consumer

PR #736, opened 2026-01-15
Extend OpenAPI document with user-generated content management endpoints

f-galland:
### Description
This PR adds an OpenAPI definition for the content manager's resource management API as defined [here](https://github.com/wazuh/internal-devel-requests/issues/3924).

### Issues Resolved
Resolves https://github.com/wazuh/internal-devel-requests/issues/3925


plugins/content-manager/resources-openapi.yml
```diff
@@ -0,0 +1,445 @@
...
+        error:
+          type: string
+          description: Short error code or identifier
+        status:
+          type: integer
+          description: HTTP status code
+        details:
+          type: object
+          additionalProperties: true
+          description: Optional structured error details
```
Jorgesnchz: I think that we have too many fields for the error response, the response should be based in [RestResponse](https://github.com/wazuh/wazuh-indexer-plugins/blob/main/plugins/content-manager/src/main/java/com/wazuh/contentmanager/rest/model/RestResponse.java) class

plugins/content-manager/resources-openapi.yml
Jorgesnchz: We could also add the header from the other openapi.yml file so make it easier to use as an imposter, allowing the usage of different protocol, ip and port ([reference](https://github.com/wazuh/wazuh-indexer-plugins/blob/37303520aed3b29c4dcd96b640862bb6f61aedcb/plugins/content-manager/openapi.yml#L12-L24))

plugins/content-manager/resources-openapi.yml
f-galland: Done in [4f45d8c](https://github.com/wazuh/wazuh-indexer-plugins/pull/736/commits/4f45d8cdc9b5ab42f8000bfa7a9ef408700c4feb)

plugins/content-manager/resources-openapi.yml
```diff
@@ -0,0 +1,445 @@
...
+              $ref: "#/components/schemas/LogtestRequest"
+      responses:
+        "200":
+          description: Logtest result
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/LogtestResponse"
+
+  /_plugins/content-manager/promote_preview:
```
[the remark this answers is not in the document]
f-galland: It is defined this way in the [specification](https://github.com/wazuh/internal-devel-requests/issues/3924).

Issue #796, opened 2026-02-09
Engine filters API


AlexRuiz7:
## Description

The Wazuh Engine can enrich the processed data using **IOCs** (Indicators of Compromise) and **filters**. IOC information will be provided by CTI, fetched and hosted by the Indexer in XDR+SIEM deployments, and consumed by the engine. Filters are user-generated.

Filters are conceptually similar to decoders but have a reduced and specialized structure. They do not perform parsing or normalization and are used only to decide whether an event should continue through the pipeline.

To accomplish that, the Indexer must provide an API to allow creating, updating, and removing this new type of resource.

Head over to https://github.com/wazuh/wazuh/issues/34146 for a list of required fields and a filter example.

Related issues:
- https://github.com/wazuh/wazuh/issues/34146
- https://github.com/wazuh/wazuh-indexer-plugins/issues/755
- https://github.com/wazuh/internal-devel-requests/issues/3899
- https://github.com/wazuh/wazuh/blob/d1d5f3ea43e9ba99427483fe634f2338ad8f87d9/docs/ref/modules/engine/spec.yaml

## Functional requirements

- The Indexer must expose an API to manage filters (stored in `.engine-filters`) using the predefined mappings (**strict mode**).
- The filters are the only resource that can be modified by the user in the standard space, not only the `draft` space.
- The Indexer must compute a hash for each filter and store it within the filter document itself.
- Policy and space hashes must be recalculated whenever a filter is created, modified, or deleted.

**Create**

```
POST /_plugins/_content_manager/filters
{
  "space": "draft",
  "resource": {
    "name": "filter/prefilter/0",
    "enabled": true,
    "metadata": {
      "description": "Default filter to allow all events (for default ruleset)",
      "author": {
        "email": "info@wazuh.com",
        "name": "Wazuh, Inc.",
        "url": "https://wazuh.com"
      }
    },
    "check": "$host.os.platform == 'ubuntu'",
    "type": "pre-filter"
  }
}
```

1. If an ID is provided in the request, return an error. The ID must be generated during creation and should not be supplied by the client.
2. If no ID is provided, generate a UUIDv4 for the resource (`document.id`).
3. Generate the creation and modification dates.
4. Call the local engine with a `POST` request to `/content/validate/resource`:    
    ```JSON
    {
      "type": "filter",
      "resource": {} // JSON Object representing the resource to be validated
    }
    ```
[...]

PR #801, opened 2026-02-10
Engine Filters API

adrianvponce:
### Description
The Wazuh Engine can enrich the processed data using IOCs (Indicators of Compromise) and filters. IOC information will be provided by CTI, fetched and hosted by the Indexer in XDR+SIEM deployments, and consumed by the engine. Filters are user-generated.

Filters are conceptually similar to decoders but have a reduced and specialized structure. They do not perform parsing or normalization and are used only to decide whether an event should continue through the pipeline.

To accomplish that, the Indexer must provide an API to allow creating, updating, and removing this new type of resource.

### Issues Resolved
Closes https://github.com/wazuh/wazuh-indexer-plugins/issues/796


f-galland:
I re-run the validation steps and the space hash is now being recalculated correctly on POST, PUT and DELETE.

f-galland (changes requested):

ecs/generator/update_module_list.sh
```diff
@@ -145,6 +145,15 @@ function map_cti_modules() {
   done
 }
 
+# ====
+# Map Engine Filter modules
+# ====
+function map_engine-filter_modules() {
```
AlexRuiz7: I'd rename to:
```suggestion
function map_engine_filter_module() {
```

AlexRuiz7 (changes requested):

PR #1027, opened 2026-04-15
Add documentation about Rule limits for detectors

### LINKED ISSUE #111: Limited number of rules for detectors
## Description

Threat Detectors can run hundreds of queries over incoming data every minute (default value). This effort comes at a performance cost. For example, we noticed findings are dropped by the alerting plugin monitors.

Related issue:
- #82  

To mitigate the problem, we are limiting the number of rules allowed per Threat Detector, which will be 100 by default. The value is NOT configurable.

We are also changing the default execution interval time for the detectors to 2 minutes, instead of 1.

## Functional requirements

- Wazuh Indexer imposes a limit of 100 rules per threat detector. This limit applies to every threat detector equally.
- Threat Detectors execution interval time is set to 2 minutes, instead of 1.

> [!NOTE]
>  The limit applies to standard threat detectors AND user-created ones.

- N/A

## Plan

- [x] Implement threat detector limit. 
- [x] Implement automated testing.
- [x] Documentation.

f-galland:
### Description
This PR modifies documentation to mention the limit on the number of rules a detector can be assigned.

### Issues Resolved
Resolves https://github.com/wazuh/wazuh-indexer-security-analytics/issues/111

### DIFF
--- a/docs/ref/modules/content-manager/architecture.md
+++ b/docs/ref/modules/content-manager/architecture.md
@@ -54,6 +54,8 @@ Interfaces with the OpenSearch Security Analytics plugin. Creates, updates, and 
 
 This design allows the same CTI resource to exist across multiple spaces without ID collisions. Association and lookup between CTI and SAP documents is performed by querying `document.id` + `source`.
 
+> **Note:** SAP enforces a maximum of 100 rules per detector. If an integration has more than 100 enabled rules, the detector creation or update request will be rejected. See [Security Analytics — Detector constraints](../security-analytics/index.md#detector-constraints) for details.
+
 ### Space Service
 
 Manages the four content spaces (standard, draft, test, custom). Routes CUD operations to the correct space partitions within system indices. Handles promotion by computing diffs between spaces in the promotion chain (Draft → Test → Custom).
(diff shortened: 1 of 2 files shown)


adrianvponce (approved):
LGTM.

QU3B1M (approved):

docs/ref/modules/content-manager/architecture.md
```diff
@@ -54,6 +54,8 @@ Interfaces with the OpenSearch Security Analytics plugin. Creates, updates, and
 
 This design allows the same CTI resource to exist across multiple spaces without ID collisions. Association and lookup between CTI and SAP documents is performed by querying `document.id` + `source`.
 
+> **Note:** SAP enforces a maximum of 100 rules per detector. If an integration has more than 100 enabled rules, the detector creation or update request will be rejected. See [Security Analytics — Detector constraints](../security-analytics/index.md#detector-constraints) for details.
```
QU3B1M, line 57: I have a doubt about this, why does it matter if an integration has more than 100 enabled rules? When we create a detector we choose the rules one by one, couldn't we just choose more than the total of 100 rules of an integration?

AlexRuiz7: This is exactly what we pretend.

PR #1078, opened 2026-04-27
Improve handling of Security Analytics responses

### LINKED ISSUE #143: Improve API response on "unknown WCS fields" validation error
## Description

When creating a rule that references fields not present in the Wazuh Common Schema (WCS), the plugin returns a `500 Internal Server Error` with an unclear message:

```json
{
  "message": "Error in Security Analytics. {\"SigmaError\":\"Unknown WCS fields in detection: [proceso.ejecutable]\"}",
  "status": 500
}
```

This is a client-side validation error (the user provided invalid field names), so it should return `400 Bad Request` The error message should also be clearer and not expose internal exception class names like `SigmaError`

## Functional requirements

- Return HTTP `400 Bad Request` instead of `500 Internal Server Error` when a rule references fields that are not part of the WCS
- Provide a clear, user-friendly error message that:
  - Explicitly states which fields are not recognized in the WCS
  - Does not expose internal exception types (`SigmaError`)
  - Example: `"Rule cannot be created. The following fields are not part of the Wazuh Common Schema (WCS): [proceso.ejecutable]"`

## Plan

- [x] Identify where the `SigmaError` for unknown WCS fields is thrown and where it is caught/propagated
- [x] Wrap the error as `OpenSearchStatusException(message, RestStatus.BAD_REQUEST)`
- [x] Reformat the error message to be user-friendly (remove `SigmaError` wrapper, mention WCS explicitly)
- [x] Add/update unit tests to verify the 400 status and message format for unknown WCS fields
- [x] Verify that other `SigmaError` cases are not affected

Jorgesnchz:
### Description
This PR creates a new exception class `SecurityAnalyticsClientException` to manage correctly the SAP errors and improves the logic used to parse the messages so it is cleared than before

Validation in: https://github.com/wazuh/wazuh-indexer-security-analytics/issues/143#issuecomment-4325581669 

### Issues Resolved
Closes https://github.com/wazuh/wazuh-indexer-security-analytics/issues/143

### DIFF
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/SecurityAnalyticsServiceImpl.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/SecurityAnalyticsServiceImpl.java
@@ -183,7 +183,7 @@ public class SecurityAnalyticsServiceImpl implements SecurityAnalyticsService {
 
     @Override
     public void upsertRule(JsonNode doc, Space space, Method method)
-            throws SecurityAnalyticsClientException {
+            throws SecurityAnalyticsException {
         if (!doc.has(Constants.KEY_ID)) {
             log.error(Constants.E_LOG_MISSING_FIELD, Constants.KEY_ID);
             return;
@@ -231,9 +231,10 @@ public class SecurityAnalyticsServiceImpl implements SecurityAnalyticsService {
                         .actionGet();
             }
         } catch (Exception e) {
-            String msg = e.getMessage() != null ? e.getMessage() : "Unknown error";
-            if (msg.contains("Wazuh Common Schema (WCS)")) {
-                throw new SecurityAnalyticsClientException(extractSapErrorMessage(msg), e);
+            String message = e.getMessage() != null ? e.getMessage() : "Unknown error";
+            if (message.contains("Wazuh Common Schema (WCS)")) {
+                throw new SecurityAnalyticsException(
+                        SecurityAnalyticsServiceImpl.extractErrorMessage(message), e);
             }
             throw e;
         }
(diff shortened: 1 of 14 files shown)


Jorgesnchz:
Since this is a fix/improvement I believe that the changelog entry is not needed, and the documentation doesn't reference this error, so I don't think we need to change anything to it

QU3B1M (approved):
LGTM

---

Validation

`request`
```JSON
{
  "integration": "{{integration_id}}",
  "resource": {
    "metadata": {
      "title": "KeePass (test)",
      "description": "Test rule",
      "references": [],
      "author": "Wazuh"
    },
    "enabled": true,
    "status": "test",
    "level": "high",
    "logsource": {
      "product": "findings-test"
    },
    "detection": {
      "condition": "selection",
      "selection": {
        "process.pepe": "KeePass.exe"
      }
    },
    "mitre": {
      "tactic": ["TA0002"],
      "technique": ["T1059"]
    }
  }
}
```
`response`
```JSON
{
    "message": "Error in Security Analytics. The following fields are not part of the Wazuh Common Schema (WCS): [process.pepe]",
    "status": 400
}
```

plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/SecurityAnalyticsServiceImpl.java
```diff
@@ -198,35 +200,66 @@ public void upsertRule(JsonNode doc, Space space, Method method) {
...
+                                        WriteRequest.RefreshPolicy.IMMEDIATE,
+                                        product,
+                                        method,
+                                        body,
+                                        true,
+                                        id,
+                                        sourceName))
+                        .actionGet();
+            }
+        } catch (Exception e) {
```
adrianvponce, line 233: Code works. However, the error is coming from a SecurityAnalyticsException. Should we have a catch entry for this specific exception?

Jorgesnchz: Though I think using another catch block for this specific exception when we try to do it the changes required are bigger than the benefits from doing this change, since `SecurityAnalyticsException` is exclusive from SAP we should need to: 
- Include it in the commons module so it can be referenced by CM and SAP
- Modify all the SAP code that refers to it

So IMO is better to leave as it is

adrianvponce: Sounds good to me. Thanks for checking.

adrianvponce (approved):
LGTM.

PR #1164, opened 2026-05-12
Modify logic to ensure token is loaded after restart

Jorgesnchz:
### Description
This PR adds a method that ensures the token is always loaded when it is needed, avoiding the error we were facing where we were losing the token value after restarting

### Issues Resolved
Closes IDR4797


AlexRuiz7 (approved):
After restart

```json
{
  "message": {
    "plan": {
      "name": "Pro Plan",
      "is_public": false
    },
    "is_registered": true
  },
  "status": 200
}
```

PR #1176, opened 2026-05-15
Enhance security of `.wazuh-internal-state` index

Jorgesnchz:
### Description
This PR changes how we were accessing to the internal index where the content is stored, also adds a guard that makes the plugin completely useless in case the index is not added to the `system_indices` list

### Issues Resolved
IDR4833


f-galland (approved):
LGTM!

ferooo23 (approved):
LGTM!

AlexRuiz7 (changes requested):
API response message is OK, but the response code isn't. Should not be a 500 Server Error.

AlexRuiz7 (approved):
Changed response code to 412 in 48460581ba6faa630c9eb31f02de415f861338d7

Image

PR #1210, opened 2026-05-25
Change threat.enrichments from nested to object and remove root-level enrichments

### LINKED ISSUE #1209: Enrichments dataset as unnested object
## Description

Currently, the field `threat.enrichments` and `wazuh.threat.enrichments` is an array of nested objects. That causes problems in the Discover section of the UI as this data type is unrecognized by the Dashboard. Not only that, but that also causes big trouble to build visualizations. 

Image

We have verified that changing that field type to `object` solves both issues:

- the Dashboard correctly detects the field as a nested object.
- data can be exploited through visualizations 

Image

Finally, we also noticed that the `enrichements` dataset is present at the root level of the stateless Wazuh Common Schema, but as far as we know, those fields are currently of no use, so we can get rid of them completely. 

Image

## Functional requirements

- No `enrichments` at root level.
- `threat.enrichments` and `wazuh.threat.enrichments` as object types.

- Adhere to the WCS tooling and guidelines. 

## Plan

- [x] Remove `enrichments` from the root level.
- [x] Change data type of `threat.enrichments` and `wazuh.threat.enrichments` fields to `object`.

ferooo23:
### Description
Changes `threat.enrichments` and `wazuh.threat.enrichments` field type from `nested` to `object`. Removes the unused root-level `enrichments` field set.

### Issues Resolved
Resolves #1209


AlexRuiz7 (approved):
LGTM

Image

PR #1232, opened 2026-06-01
Implement performance metrics tooling

### LINKED ISSUE #1590: Wazuh Indexer 5.0.0 performance metrics
## Description

The Wazuh Indexer acquires in 5.x many of the functionalities of the product that were in the Wazuh Manager in the 4.x baseline.

Most of these features have been re-implemented completely in Java, while others are completely new and did not exist before.

We are completely at a lost at the moment about Wazuh Indexer performance, as we have not carried out any performance test yet. We know, by our daily usage, that it requires way more resources than in 4.x, which is expected as per the reasons mentioned beforehand, but we don't know how much more hardware utilization it needs in a measurable unit compared to 4.x.

This is key to:

1. document minimum and recommended hardware requirements for 5.x.
2. look for improvements and measure their impact.

> _You cannot improve what you don't measure._

## Test Scenario

- AIO environment + 2 agents (beta2) sending events in a loop (FIM, Logcollector).
- 16 GB RAM, 8 CPU cores.
- Test duration: 60 minutes.
- Measures: every minute.

## Plan

- [x] Look for tools to measure resource usage on OpenSearch clusters: [OpenSearch Performance Analyzer](https://github.com/opensearch-project/performance-analyzer), [OpenSearch Benchmark](https://github.com/opensearch-project/opensearch-benchmark), [Rally](https://github.com/elastic/rally)
- [x] Detail a performance testing plan.
- [x] Implement such plan
- [x] Deliverable: performance metrics report

QU3B1M:
### Description

This PR adds a self-contained performance testing tool under `tools/performance/` to measure Wazuh Indexer performance and catch regressions across versions. It provides two complementary scenarios that share one measurement and analysis layer.

> [!NOTE]
> This tool has **its own Vagrantfile** because it requires slightly different environment than the `test-cluster`. It was made this way in order to avoid adding even more complexity to the test-cluster's Vagrantfile/tooling and making the tool "self-contained".

**Scenarios**
- **real-world**: an all-in-one node (manager, indexer, dashboard) plus two agents running FIM and Logcollector load loops. Use it for hardware sizing under realistic SIEM ingest.
- **isolated**: a single standalone indexer plus a monitor host running OpenSearch Benchmark, watched from boot via node_exporter, Prometheus, and Grafana. Use it for indexer peak throughput and cold-start cost.

**How to run**
- **One-liner (Vagrant)**: `./run.sh --scenario real-world` or `./run.sh --scenario isolated`. A single entrypoint owns the lifecycle (provision, measure, teardown).
- **Manual**: the same guest-side scripts run directly on bare hosts such as AWS EC2 or CI, using each instance private IP for inter-host traffic.
- Versions are selected as MAJOR.MINOR (for example `--version 5.0` or `--version 4.14`), and the latest patch of that line is installed. Each run lands in `runs/<scenario>-<version>/`, named after the actual installed version, so runs never overwrite each other and stay comparable.

**What it measures (per minute)**
- Host: total CPU, load, RAM, swap, disk usage, disk read/write rate, plus a per-process split across wazuh-indexer, wazuh-manager, and wazuh-dashboard.
- Indexer: ingest and query rate, per-operation index and query latency, indexing pressure, write and search thread-pool queue depth and rejections per second, JVM heap, GC count and time, segments, merge/refresh/flush, store size, and doc count.

**Analysis outputs**
- `report.md`: labeled average and peak summary for a single run.
- `compare.md`: side-by-side diff across two or more runs, for example 4.x vs 5.x.
- `timeline.png`: overlaid per-minute charts that show when spikes happen (cold start, GC, ingest dips).
- Isolated runs also auto-provision a Grafana "Host Overview" dashboard so the from-boot cold-start timeline is visible live.

**Components**
- `run.sh` and `analyze.sh`: top-level entrypoints.
- `metrics/sampler.py`: per-minute host and indexer sampler that emits CSV and NDJSON.
- `analyze/`: report, compare, and plot generators.
- `scripts/`: install and measurement scripts (setup-aio, setup-indexer, setup-agent, run-scenario, agent-load).
- `benchmark/`: OpenSearch Benchmark workload and corpus generator built from the real WCS event generator and index template.
- `monitoring/`: node_exporter, Prometheus, and Grafana stack for the isolated scenario.
- `vagrant/`: Vagrantfile, shared helpers, and per-scenario measurement runners.

**Teardown behavior**
- real-world is torn down automatically on success.
- isolated leaves its VMs up after a successful run so Grafana and Prometheus stay reachable. Tear them down with `./run.sh --scenario isolated --destroy` or by starting another run, which cleans up leftovers first.

### Issues Resolved
Closes https://github.com/wazuh/wazuh-indexer/issues/1590


Jorgesnchz (approved):
Tooling looks good, we will now be able to perform performance checks, as mentioned by @QU3B1M on a meet I think it could be a great idea to create a GHA workflow that runs this tools so we can see how the performance of the indexer evolves

The results are similar to the ones we got from QA team analysis

ana-rm:
LGTM the tool is well-structured, self-contained, and covers the requirements from the issue.

Minor suggestion:
Right now the tool only resolves versions from two sources: the latest 5.x staging nightly and the latest GA patch of a 4.x line. It would be useful to also support custom package URLs so that arbitrary builds ( feature branches, CI artifacts, local packages ) can be benchmarked without needing to publish them first.

ana-rm (approved):

PR #1242, opened 2026-06-08
Add new fields to metrics comms template for wazuh-remoted stats

### LINKED ISSUE #1235: Add new properties to metrics-comms.json for wazuh-remoted stats
# Description

In `wazuh/wazuh`, issue [#36336](https://github.com/wazuh/wazuh/issues/36336) fixes a bug in `wazuh-remoted` stats where `messages.received_breakdown.event` always reported `0`, and extends the classification of received messages. The C-side fix landed as [#36456](https://github.com/wazuh/wazuh/pull/36456) and the framework-side follow-up that updates the metrics shipper is in [#36740](https://github.com/wazuh/wazuh/pull/36740).

As part of that work, the `getstats` API of `wazuh-remoted` now exposes three additional counters under `metrics.messages.received_breakdown`. `metrics_snapshot.py` picks them up and ships them to the `wazuh-metrics-comms-*` data stream.

Because the current template uses `"dynamic": "strict"`, the indexer will reject documents containing the new fields until the mapping is updated.

## What we need

Three new `long` properties at the root of the mapping, in `plugins/setup/src/main/resources/templates/streams/metrics-comms.json`:

- `events_failed.total`
- `states.total`
- `upgrade_ack.total`

### `events.total`

Total number of agent events that `wazuh-remoted` accepted and queued for delivery to `wazuh-analysisd`. These are the regular, stateless event payloads that drive rule-based detection (syslog data, audit records, security events, and similar). Messages classified as control, inventory-sync, upgrade-ack, ping, or legacy DBSYNC are counted under their own fields and do not contribute here.

### `events_failed.total`

Total number of agent events that `wazuh-remoted` could not deliver to `wazuh-analysisd`. This includes events rejected by the internal batch queue between `wazuh-remoted` and `wazuh-analysisd` (for example when the queue is full or an agent has exceeded its per-agent cap) and events dropped by the HTTP dispatcher when the connection to `wazuh-analysisd` is unavailable or the POST fails. Note that an event dropped at dispatch time will also have been counted in `events.total` when it was first accepted into the batch queue; `events.total` reflects events that reached the batch queue, not events successfully delivered.

This is different from `messages.received_breakdown.discarded`. `discarded` counts messages that are dropped at network ingress, before classification, because the inbound queue inside `wazuh-remoted` is full. `events_failed` covers messages that have already been classified as events but are then lost on the way to `wazuh-analysisd`.

### `states.total`

Total number of inventory-synchronization messages received from agents and successfully forwarded to the `inventory-states` router provider. These are the binary flatbuffer payloads produced by the inventory modules, and on the wire they carry the header `s:{module}:` (for example `s:syscollector:` or `s:fim:`).

### `upgrade_ack.total`

Total number of upgrade-acknowledgement messages received from agents and forwarded to the `upgrade_notifications` router provider. The wire header is `u:upgrade_module:` and the body is a JSON status report describing the result of an agent-side upgrade operation.

Jorgesnchz:
### Description

This PR adds 3 new fields to the template `metrics-comms.json` as requested and accepted in the issue:
- `events.failed.total`
- `messages.states.total`
- `messages.upgrades.total`

### Issues Resolved
Closes #1235

### DIFF
--- a/plugins/setup/src/main/resources/templates/streams/metrics-comms.json
+++ b/plugins/setup/src/main/resources/templates/streams/metrics-comms.json
@@ -45,6 +45,13 @@
         },
         "events": {
           "properties": {
+            "failed": {
+              "properties": {
+                "total": {
+                  "type": "long"
+                }
+              }
+            },
             "total": {
               "type": "long"
             }
@@ -87,8 +94,22 @@
                 }
               }
             },
+            "states": {
+              "properties": {
+                "total": {
+                  "type": "long"
+                }
+              }
+            },
             "total": {
               "type": "long"
+            },
+            "upgrades": {
+              "properties": {
+                "total": {
+                  "type": "long"
+                }
+              }
             }
           }
         },
--- a/wcs/stateless/metrics/comms/docs/fields.csv
+++ b/wcs/stateless/metrics/comms/docs/fields.csv
@@ -2,8 +2,8 @@ ECS_Version,Indexed,Field_Set,Field,Type,Level,Normalization,Example,Description
 9.1.0,true,base,@timestamp,date,core,,2016-05-23T08:05:34.853Z,Date/time when the event originated.
 9.1.0,true,discarded,discarded.total,long,custom,,0,Cumulative number of discarded messages.
 9.1.0,true,event,event.module,keyword,core,,apache,Name of the module this data is coming from.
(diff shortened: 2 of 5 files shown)


Jorgesnchz:
I have not added a changelog entry since the schema is changing a lot of times, and we have not yet defined the final version, so I believe that a changelog entry is not needed here

AlexRuiz7 (approved):

PR #1361, opened 2026-07-07
Pin workflow references to the release branch instead of main

ferooo23:
### Description
Fixes CI workflows that call reusable workflows from other repos using `@main`, which breaks once a branch diverges from main . Manually pins the current `@main` references in this branch to `5.0.0`, and adds `workflow_refs_sync.sh` (called by the repository bumper on every bump) so that, any `@main` reference to another repo gets automatically pinned to the current branch whenever a release branch is bumped.

### Issues Resolved
Resolves IDR5387


ferooo23:
Replaced by: https://github.com/wazuh/wazuh-indexer-plugins/pull/1372

PR #1377, opened 2026-07-10
Fix promotion changes

### LINKED ISSUE #1373: [BUG] Promotion changes on fresh install
**Describe the bug**

On a fresh installation of Wazuh, the promotion preview endpoints returns changes between the draft and the test spaces. As a result, the new UI banner to notify the user about unpromoted changes shows up even if the user has not made any changes yet.

This is caused by a difference in the `enabled` field of these 2 spaces. In the draft space, it's true to save the user from an additional step tp enable the policy, but in the test space, it is set to false to prevent the Engine from reading this empty space.

To address this issue we ne need to resolve this inconsistency, by checking if setting `enabled:true` to every space policy has any negative impact in the Wazuh manager's performance.

**To Reproduce**
1. Go to Security Analytics

**Expected behavior**
No unpromoted changes on a fresh installation.

**Plugins**
Content Manager.

**Screenshots**

Image

Image

**Host/Environment (please complete the following information):**
 - 5.0.0 latest

**Additional context**
Add any other context about the problem her

ana-rm:
### Description

Fixes an inconsistency where policies were created with `enabled: true` for the **draft** space but `enabled: false` for every other space (test, custom, standard). Because the promote preview compares the draft and test policies, this mismatch alone was reported as an "unpromoted change", so the dashboard showed up even on a fresh installation where the user hadn't made any changes.

- `SpaceService.initializeSpace()` now sets `enabled: true` for all spaces by default, so draft and test start out identical.
- Updated `SpaceServiceTests` accordingly: the test that asserted `enabled:false` for non-draft spaces now asserts `enabled:true`.

### Issues Resolved
Closes #1373

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -239,6 +239,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fix race condition between Content Manager and Setup plugin initialization [(#1262)](https://github.com/wazuh/wazuh-indexer-plugins/pull/1262)
 - Fix invalid action name warning messages [(#1313)](https://github.com/wazuh/wazuh-indexer-plugins/pull/1313)
 - Fix consumer offset left behind the remote head after a plan-change content swap [(#1343)](https://github.com/wazuh/wazuh-indexer-plugins/pull/1343)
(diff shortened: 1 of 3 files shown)


ana-rm:
## Validations

- Before changes
image

- After changes
image

All spaces enabled:
image
image
image
image

AlexRuiz7 (approved):

PR #1411, opened 2026-07-28
Preserve the user's enabled choice on integrations

### LINKED ISSUE #1403: [BUG] User's enabled state is lost on CTI updates
**Describe the bug**

The `standard` space is rebuilt from CTI on every content update, and everything the user changed there is lost. This was reported as one field, an integration's `enabled`, and turned out to be four things:

| What the user changes | What happens on the next update |
| --- | --- |
| An integration's `enabled` | reverted to CTI's value |
| Its detector's Start/Stop | the detector is turned back on |
| A filter created in `standard` | deleted and never restored |
| The `standard` policy's settings | reverted |

The original case: Integration's mode #1356 established that `PUT /integrations` only allows modifying `enabled` on `user-managed` integrations, implying that value belongs to the user going forward, but that guarantee does not hold once CTI content updates run. Confirmed live with two independent mechanisms, a plan/subscription change and a full resync. In both, the `enabled` value the user set is discarded and overwritten with whatever CTI sends.

The other three were found while investigating it, with evidence in
[this comment](https://github.com/wazuh/wazuh-indexer-plugins/issues/1403#issuecomment-5190504630).

**To Reproduce**

*Scenario 1: plan/subscription change:*
1. Toggle a real `standard`/`user-managed` integration's `enabled` via `PUT /_plugins/_content_manager/integrations/{id}`.
2. Register a different subscription (`POST /_plugins/_content_manager/subscription` with a different access token).
3. `POST /_plugins/_content_manager/update`.
4. Search the integration by title afterward.
5. See error, `enabled` reverted to CTI's value, under a different `document.id`.

*Scenario 2: full resync:*
1. Toggle a real `standard`/`user-managed` integration's `enabled`.
2. `POST /.wazuh-cti-consumers/_delete_by_query` with `match_all` (simulates lost consumer tracking).
3. `POST /_plugins/_content_manager/update` (forces `local_offset = 0`, full reinitialization).
4. Re-check the same integration.
5. See error: `enabled` reverted to CTI's value.

The same two scenarios reproduce the other three losses: stop a detector, create a filter in `standard`, or change the policy's settings before running either one.

**What is the expected behavior?**

Everything the user decides in the `standard` space survives every path that rebuilds it: incremental updates, full resynchronisations and subscription plan changes, with no silent state loss.

The implemented design is described in [this comment](https://github.com/wazuh/wazuh-indexer-plugins/issues/1403#issuecomment-5252043353): a single registry document, `wazuh-user-overrides`, that records those decisions and is re-applied after every synchronization.

Related: #1356 (introduced the `mode`/`enabled` user-control model).

ferooo23:
## Description

A user's enable/disable choice on a `standard`/`user-managed` integration was lost whenever content arrived from CTI, because `document.enabled` had two writers  , CTI on every content update, and the user through `PUT /integrations/{id}` , and the last write won.

Closes #1403

## Proposed Changes

Each field gets a single writer:

| Field | Written by | Meaning |
| --- | --- | --- |
| `document.cti_enabled` | CTI | the state CTI publishes |
| `document.user_enabled` | Content Manager | the user's explicit choice; **absent** means never set |
| `document.enabled` | Content Manager | the effective state, resolved as `user_enabled ?? cti_enabled` |

The rule lives in one class, `IntegrationEnabledResolver`, and runs at ingest and on update — never on read. Once the user sets a value, it wins over every later CTI update.

It is applied on all five paths that rebuild integration documents: incremental CTI patches, full resynchronisation, subscription plan change (shadow index swap), rollback to a local/stable snapshot, and a failed override read (which now aborts the snapshot instead of silently rebuilding every integration with CTI's values).

Overrides are matched by `document.metadata.title`, not `document.id`, because CTI reassigns the document id between catalog generations.

**API change:** the client now sends `user_enabled` instead of `enabled` when creating or updating an integration. `enabled` is derived and no longer accepted from a request body; neither is `cti_enabled`. On create, `user_enabled` is optional and defaults to `true`, exactly as `enabled` did.

### Results and Evidence

Validated on a live cluster, with every request and response recorded here:
**[Validation comment](https://github.com/wazuh/wazuh-indexer-plugins/pull/1411#issuecomment-5120377948)**

It covers the nine scenarios end to end — the toggle and the request contract,
a full CTI resynchronisation, the rollback to the stable snapshot, a plan change
where CTI reassigns every `document.id`, and create/update/promotion across
`draft`, `test` and `custom`. Seven of them are automated by the script attached
to that comment; the two that touch the host firewall and the cluster's
subscription come with copy-paste runbooks.

### Artifacts Affected

None.

### Configuration Changes

None. Two boolean fields are added to the `wazuh-threatintel-integrations` mapping; they apply on the next index creation, and until then are stored but not queryable, with no behaviour change.

### Documentation Updates

`openapi.yml`, `docs/dev/plugins/content-manager.md` (new "Integration `enabled` resolution" section) and `CHANGELOG.md`.

### Tests Introduced

`IntegrationEnabledResolverTests` (11) and `TransportUpdateIntegrationActionTests` (9) are new; 5 added to `ContentIndexTests`, 6 to `SnapshotServiceImplTests` and 2 to `ConsumerIocServiceTests`, plus REST coverage in `IntegrationCUDIT` and `test_resources_lifecycle.py`. Every new test was verified to fail without the corresponding production change.

The three in `ConsumerIocServiceTests` and `SnapshotServiceImplTests` that guard the capture ordering deserve a mention: they assert that the overrides are read **before** the standard space is wiped, and that a failed read aborts without deleting anything. A unit test on `initialize()` alone cannot catch that — the wipe happens in the caller — which is exactly how the first version of this fix passed its tests and still lost every override on a live cluster.

## Review Checklist

- [ ] Code changes reviewed

### Notes for reviewers

- **Must land with the dashboard change** ([wazuh-dashboard-security-analytics#426](https://github.com/wazuh/wazuh-dashboard-security-analytics/pull/426)). Until the dashboard sends `user_enabled`, the toggle returns `400`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -170,6 +170,7 @@
 - [BUG] Missing threat-intel indices [(#1362)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1362)
 - Update the Content Manager OpenAPI (`openapi.yml`) to match the current API [(#1353)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1353)
 - [BUG] Vulnerability Detection empty for all agents after a transient snapshot bootstrap failure [(#1383)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1383)
+- [BUG] User's enabled state is lost on CTI updates [(#1403)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1403)
 
 ## Prior versions
 - []()
--- a/docs/dev/plugins/content-manager.md
+++ b/docs/dev/plugins/content-manager.md
@@ -379,9 +379,13 @@ are present on the document in every space (`standard`, `draft`, `test`, `custom
 
 The resolution rule is `enabled = user_enabled ?? cti_enabled`, evaluated when content is
 ingested from CTI and whenever the client creates or updates an integration — never on read.
-Once a user has set a value it wins over every subsequent CTI update, including full
-resynchronisations and subscription plan changes. In a plan change CTI reassigns `document.id`,
-so overrides are carried across by `document.metadata.title`.
+Once a user has set a value it wins over every subsequent CTI update, including incremental
+patches and full resynchronisations.
+
+Overrides are matched by `document.id`. A full resynchronisation deletes the standard-space
+documents before repopulating them, so they are read and held in memory beforehand and re-applied
+as each document is rebuilt. **Known limitation:** a subscription plan change may republish an
+integration under a different `document.id`, and the override is not carried across in that case.
 
 Until CTI publishes `cti_enabled`, it still writes `document.enabled` directly; in that case the
 resolution leaves `enabled` as received, so the user's override still takes precedence and
@@ -978,7 +982,7 @@ The plugin includes integration tests defined in the `tests/content-manager` dir
 | Resource / operation   | Scenario count | Covers                                                                                                                                                                                                                                                                                                                              |
 | ---------------------- | -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Integrations: create   | 9              | Success; duplicate title; missing title/author/category; explicit `id` in resource; missing resource object; empty body; no authentication                                                                                                                                                                                          |
-| Integrations: update   | 13             | Success; title collision with an existing draft integration; missing required fields; not found; invalid UUID; `id` in request body; attempting to add/remove dependency lists; no authentication; protected integration rejected; toggling `user_enabled` on a user-managed integration in the standard space derives `enabled`; user-managed standard update changes only `user_enabled` (other fields preserved); protected standard integration rejected; toggling `user_enabled` leaves the CTI-owned `detector` block immutable; a user's `enabled` override survives a document republished under a new `document.id` with the same title (resync / plan change) |
+| Integrations: update   | 12             | Success; title collision with an existing draft integration; missing required fields; not found; invalid UUID; `id` in request body; attempting to add/remove dependency lists; no authentication; protected integration rejected; toggling `user_enabled` on a user-managed integration in the standard space derives `enabled`; user-managed standard update changes only `user_enabled` (other fields preserved); protected standard integration rejected; toggling `user_enabled` leaves the CTI-owned `detector` block immutable |
 | Integrations: delete   | 7              | Success (no attached resources); has attached resources; not found; invalid UUID; missing ID; not in draft space; no authentication                                                                                                                                                                                                 |
 | Decoders: create       | 7              | Success; missing integration reference; explicit `id` in resource; integration not in draft space; missing resource object; empty body; no authentication                                                                                                                                                                           |
 | Decoders: update       | 7              | Success; not found; invalid UUID; not in draft space; missing resource object; empty body; no authentication                                                                                                                                                                                                                        |
--- a/docs/dev/plugins/content-manager.md
+++ b/docs/dev/plugins/content-manager.md
@@ -367,8 +367,9 @@ When a `standard` integration's effective `enabled` changes, its related Securit
 ### Integration `enabled` resolution
 
 Every integration document carries three boolean fields, each with a single writer. All three
-are present the same way in every space (`standard`, `draft`, `test`, `custom`), exactly like
-`mode`:
+are present on the document in every space (`standard`, `draft`, `test`, `custom`), exactly like
+`mode` — which is a property of the stored document, not of what a client may send. See
+[Where the client can write](#where-the-client-can-write) below for that distinction:
 
 | Field | Written by | Meaning |
 | --- | --- | --- |
@@ -386,21 +387,29 @@ Until CTI publishes `cti_enabled`, it still writes `document.enabled` directly; 
 resolution leaves `enabled` as received, so the user's override still takes precedence and
 current behaviour is otherwise unchanged.
 
-The client always sends `user_enabled`, uniformly across spaces, so the request shape is
-identical regardless of where an integration lives. The parallel with `mode` is that the field
-is present in every space even where its value is predictable — but unlike `mode`, which is
-server-set and never accepted from a request, `user_enabled` is the one field the client does
-send:
+#### Where the client can write
 
-- On create, `user_enabled` is optional and defaults to `true` when absent or `null`, the same
-  way `enabled` used to be defaulted.
-- On a `draft`-space update, `user_enabled` is required, alongside `category` and the metadata
-  fields.
-- On a `standard`-space update, `user_enabled` is the only field read; every other field the
-  client sends is ignored and the stored document is preserved untouched.
+The API exposes integration writes in two spaces only, so those are the only places a client
+ever sends `user_enabled`:
 
-`enabled` and `cti_enabled` are server-managed in every space: both are stripped from request
-bodies and cannot be set by a client, exactly like `mode`.
+| Operation | Spaces accepted | `user_enabled` |
+| --- | --- | --- |
+| `POST /integrations` | `draft` | Optional; defaults to `true` when absent or `null`, the same way `enabled` used to be defaulted. |
+| `PUT /integrations/{id}` | `draft`, `standard` | Required. |
+
+- On a `draft`-space update it is required alongside `category` and the metadata fields.
+- On a `standard`-space update it is the only field read; every other field the client sends is
+  ignored and the stored document is preserved untouched.
+
+`test` and `custom` have no write endpoint for integrations — `PUT` rejects them with
+`400 RESOURCE_SPACE_MISMATCH` (`TransportUpdateIntegrationAction.validSpaces`). Documents reach
+those spaces only through promotion (`draft → test → custom`), which copies `user_enabled` and
+`enabled` along with the rest of the document. Changing the enabled state of an integration that
+already lives in `custom` therefore means editing its `draft` copy and promoting the change
+through the chain, not calling `PUT` on the `custom` document.
+
+Wherever a client can write, `enabled` and `cti_enabled` are stripped from the request body and
+cannot be set, exactly like `mode`.
 
 "Absent `user_enabled` means the user never decided" still holds, but since create always
 defaults the field and update always requires it, the absent case now only arises for a
--- a/docs/dev/plugins/content-manager.md
+++ b/docs/dev/plugins/content-manager.md
@@ -357,7 +357,7 @@ The value is set by whoever produces the content: CTI content carries its own `m
 | Integration | Space | `PUT /integrations/{id}` |
 | --- | --- | --- |
 | `protected` | any | Rejected with `400 Bad Request`. |
-| `user-managed` | `draft` | Fully editable (metadata, category, `enabled`). |
+| `user-managed` | `draft` | Fully editable (metadata, category, `user_enabled`). |
 | `user-managed` | `standard` | Only `user_enabled` can change; every other field, including `enabled`, is preserved from the stored document. |
 
 When a `standard` integration's effective `enabled` changes, its related Security Analytics **detector is disabled/enabled in lockstep** as part of the same update flow. The detector shares the integration's document id, so the Content Manager calls the Security Analytics `WSetDetectorEnabledAction` (`setDetectorEnabled(id, enabled)`) to flip **only** the `enabled` flag on the existing detector, preserving its inputs, triggers and monitors. If the detector sync fails the whole update is aborted, so the two never drift.
@@ -366,30 +366,47 @@ When a `standard` integration's effective `enabled` changes, its related Securit
 
 ### Integration `enabled` resolution
 
-A standard integration carries three boolean fields, each with a single writer:
+Every integration document carries three boolean fields, each with a single writer. All three
+are present the same way in every space (`standard`, `draft`, `test`, `custom`), exactly like
+`mode`:
 
 | Field | Written by | Meaning |
 | --- | --- | --- |
-| `document.cti_enabled` | CTI | The state CTI publishes. |
-| `document.user_enabled` | Content Manager | The user's explicit choice. **Absent** means the user never set it. |
-| `document.enabled` | Content Manager | The effective state. Everything that reads an integration's enabled state reads this. |
+| `document.cti_enabled` | CTI | The state CTI publishes. Absent on integrations created through the REST API, since that content does not come from CTI. |
+| `document.user_enabled` | Content Manager | The user's explicit choice, set from the `user_enabled` field the client sends. |
+| `document.enabled` | Content Manager | The effective state, derived and never accepted from a request. Everything that reads an integration's enabled state reads this. |
 
 The resolution rule is `enabled = user_enabled ?? cti_enabled`, evaluated when content is
-ingested from CTI and when the user toggles the integration — never on read. Once a user has
-set a value it wins over every subsequent CTI update, including full resynchronisations and
-subscription plan changes. In a plan change CTI reassigns `document.id`, so overrides are
-carried across by `document.metadata.title`.
+ingested from CTI and whenever the client creates or updates an integration — never on read.
+Once a user has set a value it wins over every subsequent CTI update, including full
+resynchronisations and subscription plan changes. In a plan change CTI reassigns `document.id`,
+so overrides are carried across by `document.metadata.title`.
 
 Until CTI publishes `cti_enabled`, it still writes `document.enabled` directly; in that case the
 resolution leaves `enabled` as received, so the user's override still takes precedence and
 current behaviour is otherwise unchanged.
 
-On a `standard`-space update, `user_enabled` is the field the client sends — it is the only
-mutable field there, and `enabled` is not accepted from the request. `cti_enabled` is
-server-managed in every space: it is stripped from request bodies and cannot be set by a
-client, exactly like `mode`. `user_enabled` has no meaning outside the `standard` space and is
-likewise stripped from `draft`-space requests, including integration creation (which always
-starts in `draft`).
+The client always sends `user_enabled`, uniformly across spaces, so the request shape is
+identical regardless of where an integration lives. The parallel with `mode` is that the field
+is present in every space even where its value is predictable — but unlike `mode`, which is
+server-set and never accepted from a request, `user_enabled` is the one field the client does
+send:
+
+- On create, `user_enabled` is optional and defaults to `true` when absent or `null`, the same
+  way `enabled` used to be defaulted.
+- On a `draft`-space update, `user_enabled` is required, alongside `category` and the metadata
+  fields.
+- On a `standard`-space update, `user_enabled` is the only field read; every other field the
+  client sends is ignored and the stored document is preserved untouched.
+
+`enabled` and `cti_enabled` are server-managed in every space: both are stripped from request
+bodies and cannot be set by a client, exactly like `mode`.
+
+"Absent `user_enabled` means the user never decided" still holds, but since create always
+defaults the field and update always requires it, the absent case now only arises for a
+CTI-published `standard` integration the user has not yet touched. Readers should still test for
+presence rather than truthiness — an explicit `false` is a real decision — but should not expect
+the absent case on integrations created through the API.
 
 ---
 
--- a/docs/dev/plugins/content-manager.md
+++ b/docs/dev/plugins/content-manager.md
@@ -358,9 +358,38 @@ The value is set by whoever produces the content: CTI content carries its own `m
 | --- | --- | --- |
 | `protected` | any | Rejected with `400 Bad Request`. |
 | `user-managed` | `draft` | Fully editable (metadata, category, `enabled`). |
-| `user-managed` | `standard` | Only `enabled` can change; every other field is preserved from the stored document. |
+| `user-managed` | `standard` | Only `user_enabled` can change; every other field, including `enabled`, is preserved from the stored document. |
 
-When a `standard` integration's `enabled` is toggled, its related Security Analytics **detector is disabled/enabled in lockstep** as part of the same update flow. The detector shares the integration's document id, so the Content Manager calls the Security Analytics `WSetDetectorEnabledAction` (`setDetectorEnabled(id, enabled)`) to flip **only** the `enabled` flag on the existing detector, preserving its inputs, triggers and monitors. If the detector sync fails the whole update is aborted, so the two never drift.
+When a `standard` integration's effective `enabled` changes, its related Security Analytics **detector is disabled/enabled in lockstep** as part of the same update flow. The detector shares the integration's document id, so the Content Manager calls the Security Analytics `WSetDetectorEnabledAction` (`setDetectorEnabled(id, enabled)`) to flip **only** the `enabled` flag on the existing detector, preserving its inputs, triggers and monitors. If the detector sync fails the whole update is aborted, so the two never drift.
+
+---
+
+### Integration `enabled` resolution
+
+A standard integration carries three boolean fields, each with a single writer:
+
+| Field | Written by | Meaning |
+| --- | --- | --- |
+| `document.cti_enabled` | CTI | The state CTI publishes. |
+| `document.user_enabled` | Content Manager | The user's explicit choice. **Absent** means the user never set it. |
+| `document.enabled` | Content Manager | The effective state. Everything that reads an integration's enabled state reads this. |
+
+The resolution rule is `enabled = user_enabled ?? cti_enabled`, evaluated when content is
+ingested from CTI and when the user toggles the integration — never on read. Once a user has
+set a value it wins over every subsequent CTI update, including full resynchronisations and
+subscription plan changes. In a plan change CTI reassigns `document.id`, so overrides are
+carried across by `document.metadata.title`.
+
+Until CTI publishes `cti_enabled`, it still writes `document.enabled` directly; in that case the
+resolution leaves `enabled` as received, so the user's override still takes precedence and
+current behaviour is otherwise unchanged.
+
+On a `standard`-space update, `user_enabled` is the field the client sends — it is the only
+mutable field there, and `enabled` is not accepted from the request. `cti_enabled` is
+server-managed in every space: it is stripped from request bodies and cannot be set by a
+client, exactly like `mode`. `user_enabled` has no meaning outside the `standard` space and is
+likewise stripped from `draft`-space requests, including integration creation (which always
+starts in `draft`).
 
 ---
 
@@ -923,7 +952,7 @@ The plugin includes integration tests defined in the `tests/content-manager` dir
 | Resource / operation   | Scenario count | Covers                                                                                                                                                                                                                                                                                                                              |
 | ---------------------- | -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Integrations: create   | 9              | Success; duplicate title; missing title/author/category; explicit `id` in resource; missing resource object; empty body; no authentication                                                                                                                                                                                          |
-| Integrations: update   | 12             | Success; title collision with an existing draft integration; missing required fields; not found; invalid UUID; `id` in request body; attempting to add/remove dependency lists; no authentication; protected integration rejected; toggling `enabled` on a user-managed integration in the standard space; user-managed standard update changes only `enabled` (other fields preserved); protected standard integration rejected |
+| Integrations: update   | 13             | Success; title collision with an existing draft integration; missing required fields; not found; invalid UUID; `id` in request body; attempting to add/remove dependency lists; no authentication; protected integration rejected; toggling `user_enabled` on a user-managed integration in the standard space derives `enabled`; user-managed standard update changes only `user_enabled` (other fields preserved); protected standard integration rejected; toggling `user_enabled` leaves the CTI-owned `detector` block immutable; a user's `enabled` override survives a document republished under a new `document.id` with the same title (resync / plan change) |
 | Integrations: delete   | 7              | Success (no attached resources); has attached resources; not found; invalid UUID; missing ID; not in draft space; no authentication                                                                                                                                                                                                 |
 | Decoders: create       | 7              | Success; missing integration reference; explicit `id` in resource; integration not in draft space; missing resource object; empty body; no authentication                                                                                                                                                                           |
 | Decoders: update       | 7              | Success; not found; invalid UUID; not in draft space; missing resource object; empty body; no authentication                                                                                                                                                                                                                        |
--- a/plugins/content-manager/openapi.yml
+++ b/plugins/content-manager/openapi.yml
@@ -891,6 +891,12 @@ paths:
 
         **Note:** Integrations created through this endpoint are always stored with `mode: user-managed`.
         The `mode` field is set server-side and must NOT be included in the request body.
+
+        **Note:** `user_enabled` is optional. When omitted or `null`, it defaults to `true` — the
+        same way `enabled` used to be defaulted. The effective `enabled` is then derived by the
+        indexer as `user_enabled ?? cti_enabled`; there is no `cti_enabled` for an integration
+        created here, since this content does not come from CTI, so the derivation always yields
+        `user_enabled`. `enabled` is never accepted from the request body.
       operationId: createIntegration
       requestBody:
         required: true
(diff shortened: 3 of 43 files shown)


ferooo23:
# Proposal

## The decisions taken

**1. Three fields instead of one, each with a single writer.**

| Field | Written by | Meaning |
| --- | --- | --- |
| `document.cti_enabled` | CTI | the state CTI publishes |
| `document.user_enabled` | the user, through the API | the user's choice |
| `document.enabled` | the indexer | the effective state, resolved from the two above |

The indexer resolves `enabled` as "the user's choice if there is one, otherwise CTI's value".

**2. `document.enabled` keeps its name and meaning.** It is still the effective state, so everything that reads an integration's enabled state keeps working unchanged. What changes is that only the indexer writes it now.

**3. Integrations are created and edited with `user_enabled`, not with `enabled`.** In every space — `standard`, `draft`, `test` and `custom` — the client sends `user_enabled` and the indexer derives `enabled` from it. `enabled` is no longer accepted in a request body.

**4. The three fields are present in every space**, the same way `mode` is today: uniform even where the value is predictable, so the API behaves identically regardless of which space an integration lives in. On user-created integrations `cti_enabled` simply doesn't exist, since that content doesn't come from CTI.

**5. The user's choice wins permanently.** Once set, it survives every CTI content update: incremental updates, full resynchronisations and subscription plan changes. CTI's value is never restored automatically.

**6. If the indexer cannot read the stored user choices, the content update is aborted** rather than applied. An aborted update retries on the next cycle; an overwritten user choice would be gone for good.

## What we need from Dashboard

**Send `user_enabled` instead of `enabled` when creating or editing an integration.**

```json
PUT /_plugins/_content_manager/integrations/{id}
{ "resource": { "id": "{id}", "user_enabled": true } }
```

This applies in every space, not only to the enable/disable toggle on `standard` integrations.

Reading doesn't change: `document.enabled` still carries the effective state.

## What we need from CTI

**Publish the integration state in `document.cti_enabled` instead of `document.enabled`.**

That is what removes the root cause: `enabled` becomes a field only the indexer writes, so a content update can no longer overwrite what the user decided.

The cloud team has already confirmed this is feasible and asked us to open an IDR for it.

ferooo23:
# Validation 🟢

> [!IMPORTANT]
> **Prerequisites.** Requires a **Pro subscription** registered
> (`POST /_plugins/_content_manager/subscription`) so the CTI content syncs;
> without it there are no integrations and the validation cannot run. Wait for
> the initial CTI consumer sync to finish before starting
> (`GET /wazuh-threatintel-integrations/_count` must return `count > 0`).

> [!IMPORTANT]
> **Start from a freshly created index.** The two new fields only enter the
> mapping when `wazuh-threatintel-integrations` is created. On a cluster whose
> index predates this change they are stored but **not queryable**, so every
> search below returns nothing and the checks would pass or fail for the wrong
> reason. Either install clean, or delete the index and let the next sync
> recreate it.

> [!NOTE]
> These checks can be run automatically with the attached script:
> `IDX=https://localhost:9200 AUTH=admin:admin bash validate-issue-1403.sh`
>
> It covers tests 1, 2, 3, 6, 7, 8 and 9. Test 3 forces a full resync and takes
> several minutes; skip it with `RESYNC=0`. Tests 4 and 5 (stable-snapshot
> rollback and plan change) are **not** scripted — copy-paste runbooks for them
> are in the next block.
>
> **Download the script:** `validate-issue-1403.sh`
[validate-issue-1403.sh](https://github.com/user-attachments/files/30512547/validate-issue-1403.sh)

<details>
<summary><b>Running the two manual tests</b>: copy-paste runbooks for tests 4 and 5</summary>

Tests 4 and 5 are deliberately left out of the script because they touch state
outside the cluster's content indices — the **host firewall** and the
**subscription** — and a validation script that manipulates either and then dies
half-way leaves the machine in a worse state than it found it. They are short,
and every command is below.

Run them **after** the script, on the same environment: they reuse the overrides
it left behind.

---

## Test 4 — rollback to the stable snapshot

Makes CTI unreachable so the plugin falls back to the last known-good snapshot
on disk, then checks the overrides still land.

> [!IMPORTANT]
> The stable snapshot must exist first, otherwise there is nothing to roll back
> to and the test silently exercises another path:
> ```bash
> sudo ls -l /usr/share/wazuh-indexer/plugins/wazuh-indexer-content-manager/snapshots/ruleset.stable.zip
> ```
> If it is missing, check that the directory is owned by `wazuh-indexer` — with
> `root:root` the plugin cannot write it and only logs a `WARN`.

**1. Block the CTI host and prove the block works.**

```bash
CTI_IPS=$(getent ahostsv4 api.pre.cloud.wazuh.com | awk '{print $1}' | sort -u)
echo "$CTI_IPS"
for ip in $CTI_IPS; do
  sudo iptables -I OUTPUT -d "$ip" -p tcp --dport 443 -m comment --comment 'cti-test-block' -j REJECT
done
curl -sS --max-time 5 https://api.pre.cloud.wazuh.com/ ; echo "exit=$?"
```

Must fail in milliseconds with `exit=7`. If it answers, another IP is in play;
re-run the resolution and block that one too.

> [!NOTE]
> `iptables -j REJECT` rather than an `/etc/hosts` entry: the HTTP client keeps
> established keep-alive connections to the already-resolved address, so a
> DNS-level block has no effect until the node is restarted.

**2. Force the cycle** (Dev Tools):

```json
POST /.wazuh-cti-consumers/_delete_by_query?refresh=true
{ "query": { "match_all": {} } }
```

```json
POST /_plugins/_content_manager/update
```

A `version_conflict` on the `vulnerabilities` consumer is harmless — it only
means that consumer was mid-sync. What matters is that `ruleset` was deleted.

**3. Confirm which path ran** (wait ~30 s):

```bash
sudo grep -iE "stable|Captured" /var/log/wazuh-indexer/wazuh-cluster.log | tail -5
```

Both lines must appear, adjacent:

```
Remote snapshot failed for consumer [cti:catalog:consumer:ruleset].
  Rolling back from stable snapshot [.../ruleset.stable.zip].
Captured N integration enabled override(s) to carry across the snapshot
```

Without the first line you exercised a different path and the result does not
count.

**4. Verify BEFORE unblocking.** Unblocking triggers a fresh remote sync that
overwrites the rollback's result, and then you cannot tell which produced it.

```json
GET wazuh-threatintel-integrations/_search
{
  "size": 5,
  "query": { "exists": { "field": "document.user_enabled" } },
  "_source": ["document.metadata.title", "document.enabled", "document.user_enabled"]
}
```

**5. Unblock.**

```bash
for ip in $CTI_IPS; do
  sudo iptables -D OUTPUT -d "$ip" -p tcp --dport 443 -m comment --comment 'cti-test-block' -j REJECT
done
sudo iptables -L OUTPUT -n | grep -c cti-test-block
```

Must print `0`. Keep `$CTI_IPS` in the same shell; opening another terminal
loses it and leaves the rules in place.

---

## Test 5 — plan change (shadow swap)

Moves the cluster to a different catalog so CTI reassigns every `document.id`,
then checks the overrides still find their integrations by title.

> [!IMPORTANT]
> Two preconditions, both easy to miss:
> - `snapshots/manifest.json` must exist — it provides the free-plan resource the
>   swap targets. It ships with the full indexer package, **not** with the plugin
>   zip, so a dev environment built from the plugin alone will not have it and
>   the swap will not trigger.
> - The `ruleset` consumer must carry its Pro identity. Check it:
>   ```json
>   GET .wazuh-cti-consumers/_doc/cti:catalog:consumer:ruleset
>   ```
>   `resource` and `context` must be populated. If `resource` is `null` — which
>   happens after a rollback — delete the consumers and run one normal
>   `POST /_plugins/_content_manager/update` first, so a full initialisation
>   rewrites the identity fields.

**1. Record the starting ids.**

```json
GET wazuh-threatintel-integrations/_search
{
  "size": 5,
  "query": { "exists": { "field": "document.user_enabled" } },
  "_source": ["document.metadata.title", "document.id", "document.enabled", "document.user_enabled"]
}
```

Note the `_index` too (`...-a` or `...-b`) — it changes on the swap.

**2. Drop to the free plan.**

```json
DELETE /_plugins/_content_manager/subscription
```

```json
GET /_plugins/_content_manager/subscription
```

Must report `"name": "Free"` and `"is_registered": false` before continuing.

```json
POST /_plugins/_content_manager/update
```

**3. Wait for all three consumers to settle.**

```bash
while true; do
  RESP=$(curl -sk -u admin:admin "https://localhost:9200/.wazuh-cti-consumers/_search?size=10&_source=type,status")
  echo "$(date +%H:%M:%S) $(echo "$RESP" | grep -o '"status":"[a-z]*"' | tr '\n' ' ')"
  TOTAL=$(echo "$RESP" | grep -o '"status":"[a-z]*"' | wc -l)
  READY=$(echo "$RESP" | grep -o '"status":"ready"' | wc -l)
  if [ "$TOTAL" -eq 3 ] && [ "$READY" -eq 3 ]; then echo "=== READY ==="; break; fi
  sleep 15
done
```

**4. Confirm the swap really happened** before looking at anything else:

```json
GET .wazuh-cti-consumers/_doc/cti:catalog:consumer:ruleset
```

`context` must have changed (Pro `beta4-t0-ruleset-5` → free `beta3-t1-ruleset-5`)
and `is_public` must now be `true`. If `context` is unchanged, there was no swap.

**5. Verify.** Re-run the search from step 1 and compare: a different `_index`,
**every** `document.id` different, and the overrides still on the right titles.

**6. Restore the Pro subscription.**

```json
POST /_plugins/_content_manager/subscription
{ "access_token": "<PRO_TOKEN>" }
```

```json
POST /_plugins/_content_manager/update
```

This triggers a second swap, back to the Pro catalog — which incidentally
exercises the same path in the opposite direction.

</details>


## What we are validating

That a user's enable/disable choice on a `standard`/`user-managed` integration
survives every path that rebuilds integration documents, and that the three
fields keep a single writer each:

| Field | Written by | Meaning |
| --- | --- | --- |
| `document.cti_enabled` | CTI | the state CTI publishes |
| `document.user_enabled` | Content Manager | the user's explicit choice; **absent** means never set |
| `document.enabled` | Content Manager | the effective state, resolved as `user_enabled ?? cti_enabled` |

---
<details>
<summary><b>Environment setup</b>: expand if you are reusing a cluster that already had the plugin</summary>

Skip this on a clean install. The script does not do it for you on purpose — it
deletes indices and restarts the node, which no validation script should do to
someone else's cluster.

## 1. Drop the subscription

```json
DELETE /_plugins/_content_manager/subscription
```

## 2. Delete the content indices

```json
DELETE .wazuh*
DELETE wazuh-*
DELETE .opendistro-alerting-config
DELETE .opensearch-sap*
DELETE _data_stream/wazuh*
DELETE _index_template/*
```

Some will answer `index_not_found_exception`; that is fine, keep going.

## 3. Restart so the setup recreates everything

```bash
sudo systemctl restart wazuh-indexer
```

## 4. Re-register and sync

```json
PUT /_cluster/settings
{ "persistent": { "plugins.content_manager.wazuh_uid": "<WAZUH_UID>" } }
```

```json
POST /_plugins/_content_manager/subscription
{ "access_token": "<PRO_TOKEN>" }
```

```json
POST /_plugins/_content_manager/update
```

## 5. Confirm the mapping before running anything

```json
GET wazuh-threatintel-integrations/_mapping/field/document.user_enabled,document.cti_enabled
```

Both fields must come back. If they do not, the index was not recreated and
nothing below is meaningful.

</details>

---

<details>
<summary><b>Step-by-step in Dev Tools</b>: expand to review every request and response</summary>

# Setup: pick an integration and record its starting state

## Confirm the new fields are queryable

### REQUEST

```json
GET wazuh-threatintel-integrations/_mapping/field/document.user_enabled,document.cti_enabled
```

### REPLY

```json
{
  "wazuh-threatintel-integrations-a": {
    "mappings": {
      "document.cti_enabled":  { "full_name": "document.cti_enabled",  "mapping": { "cti_enabled":  { "type": "boolean" } } },
      "document.user_enabled": { "full_name": "document.user_enabled", "mapping": { "user_enabled": { "type": "boolean" } } }
    }
  }
}
```

## Baseline: nobody has decided anything, and CTI publishes no `cti_enabled` yet

### REQUEST

```json
GET wazuh-threatintel-integrations/_count
{ "query": { "exists": { "field": "document.user_enabled" } } }
```

### REPLY

```json
{ "count": 0 }
```

### REQUEST

```json
GET wazuh-threatintel-integrations/_count
{ "query": { "exists": { "field": "document.cti_enabled" } } }
```

### REPLY

```json
{ "count": 0 }
```

> [!NOTE]
> The second `0` is the transition state: CTI has not shipped `cti_enabled` yet,
> so it still writes `document.enabled` directly. The resolution leaves that
> value untouched when neither new field is present.

## Find a `standard` / `user-managed` integration

### REQUEST

```json
GET wazuh-threatintel-integrations/_search
{
  "size": 5,
  "query": { "bool": { "filter": [
    { "term": { "space.name": "standard" } },
    { "term": { "document.mode": "user-managed" } }
  ]}},
  "_source": ["document.id", "document.metadata.title", "document.enabled", "document.mode"]
}
```

### REPLY (truncated)

```json
{
  "hits": { "total": { "value": 108 }, "hits": [
    { "_source": { "document": {
      "mode": "user-managed",
      "metadata": { "title": "iptables" },
      "id": "4cd06a34-e133-48f3-b32e-50ea6eeae849",
      "enabled": false } } }
  ]}
}
```

> [!IMPORTANT]
> CTI ships **all 108** `user-managed` integrations with `enabled: false`. The
> 15 that arrive enabled are Wazuh core content (`protected`) and the API
> rejects any change to them. This is why the reproduction enables one rather
> than disabling it: it is the only direction where the user's value and CTI's
> can disagree with the current catalog.

Keep the id as `<INTEGRATION_ID>` and, more importantly, the **title**
(`iptables`) — that is the key the override travels by.

---

# Test 1: the user's choice is stored and `enabled` is derived from it

## Enable the integration

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/<INTEGRATION_ID>
{ "resource": { "user_enabled": true } }
```

### REPLY

```json
{ "message": "4cd06a34-e133-48f3-b32e-50ea6eeae849", "status": 200 }
```

## Key check: `enabled` flipped, derived from the override

### REQUEST

```json
GET wazuh-threatintel-integrations/_doc/<INTEGRATION_ID>
```

### REPLY (relevant fields)

```json
{
  "_version": 2,
  "_source": { "document": {
    "metadata": { "title": "iptables", "modified": "2026-07-29T11:58:07Z" },
    "mode": "user-managed",
    "enabled": true,
    "user_enabled": true
  }, "space": { "name": "standard" } }
}
```

> [!NOTE]
> **Expected:** `enabled` went from `false` to `true`, derived from
> `user_enabled`. No `cti_enabled` was invented. Note `_version: 2` — it is the
> reference point for the resync below.

---

# Test 2: `enabled` and `cti_enabled` are never accepted from a request

## A request that still sends the old field is rejected

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/<INTEGRATION_ID>
{ "resource": { "enabled": true } }
```

### REPLY

```json
{ "message": "Missing [user_enabled] field.", "status": 400 }
```

## `enabled` sent alongside `user_enabled` is ignored

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/<INTEGRATION_ID>
{ "resource": { "user_enabled": false, "enabled": true } }
```

### REPLY

```json
{ "message": "<INTEGRATION_ID>", "status": 200 }
```

### REQUEST

```json
GET wazuh-threatintel-integrations/_doc/<INTEGRATION_ID>
```

### REPLY (relevant fields)

```json
{ "_source": { "document": { "enabled": false, "user_enabled": false } } }
```

> [!NOTE]
> **Expected:** `enabled: false`. The client asked for `true`; the server
> derived `false` from `user_enabled` and discarded what was sent.

## A client cannot forge `cti_enabled`

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/<INTEGRATION_ID>
{ "resource": { "user_enabled": true, "cti_enabled": true } }
```

### REPLY

```json
{ "message": "<INTEGRATION_ID>", "status": 200 }
```

### REQUEST

```json
GET wazuh-threatintel-integrations/_doc/<INTEGRATION_ID>
```

### REPLY (relevant fields)

```json
{ "_source": { "document": { "enabled": true, "user_enabled": true } } }
```

> [!NOTE]
> **Expected:** no `cti_enabled` anywhere in the stored document. That field
> belongs to CTI and is stripped from every request body.

## A protected integration cannot be toggled at all

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/c2c959c1-d712-434e-b0c4-e84fe612729e
{ "resource": { "user_enabled": false } }
```

### REPLY

```json
{ "message": "Integration [c2c959c1-d712-434e-b0c4-e84fe612729e] is protected and cannot be modified.", "status": 400 }
```

---

# Test 3: THE BUG — the choice survives a full CTI resynchronisation

This is the scenario in the issue. Before the fix, `enabled` reverted to CTI's
value here and `user_enabled` disappeared.

## Force a full resync

### REQUEST

```json
POST /.wazuh-cti-consumers/_delete_by_query?refresh=true
{ "query": { "match_all": {} } }
```

### REPLY

```json
{ "total": 3, "deleted": 3, "failures": [] }
```

### REQUEST

```json
POST /_plugins/_content_manager/update
```

### REPLY

```json
{ "message": "The update request has been accepted for processing.", "status": 202 }
```

Wait for the `ruleset` consumer to return to `ready`:

### REQUEST

```json
GET /.wazuh-cti-consumers/_search
{ "_source": ["type", "status", "local_offset", "remote_offset"] }
```

### REPLY (relevant fields)

```json
{ "hits": { "hits": [
  { "_source": { "type": "cti:catalog:consumer:ruleset", "status": "ready",
                 "local_offset": 1239, "remote_offset": 1239 } }
]}}
```

## The log line that proves the capture happened at the right moment

```
[2026-07-29T14:57:34,436][INFO ][c.w.c.c.c.s.SnapshotServiceImpl] [indexer]
  Captured 1 integration enabled override(s) to carry across the snapshot
```

> [!IMPORTANT]
> This line is the observable difference. The overrides must be read **before**
> the standard space is wiped; reading afterwards yields an empty map that is
> indistinguishable from "the user never chose anything". If it reports
> `Captured 0` while an override exists, the capture is running too late.

## Key check: the override survived

### REQUEST

```json
GET wazuh-threatintel-integrations/_doc/<INTEGRATION_ID>
```

### REPLY (relevant fields)

```json
{
  "_version": 7,
  "_source": {
    "document": {
      "metadata": { "title": "iptables", "modified": "2026-07-21" },
      "enabled": true,
      "user_enabled": true
    },
    "offset": 1105
  }
}
```

> [!IMPORTANT]
> **Expected, and all three matter:**
> - `_version` advanced past `2` — the document really was rebuilt, so a passing
>   result cannot be an artifact of nothing having happened.
> - `metadata.modified` is back to CTI's `2026-07-21` (it was our
>   `2026-07-29T11:58:07Z`) and `offset` is CTI's — the body genuinely came from
>   the snapshot.
> - `enabled` and `user_enabled` are still `true`, against CTI's `false`.

## Several overrides at once, plus an untouched control

Repeat with three integrations enabled by the user (`iptables`, `aws-fargate`,
`netflow`) and a fourth left alone (`iis`):

### REQUEST

```json
GET wazuh-threatintel-integrations/_search
{
  "size": 4,
  "query": { "terms": { "document.metadata.title": ["iptables", "aws-fargate", "netflow", "iis"] } },
  "_source": ["document.metadata.title", "document.enabled", "document.user_enabled"]
}
```

### REPLY (relevant fields)

```
iptables      user_enabled=true   enabled=true
aws-fargate   user_enabled=true   enabled=true
netflow       user_enabled=true   enabled=true
iis           <no user_enabled>   enabled=false
```

> [!NOTE]
> **Expected:** the three overrides survive, and the control still follows CTI
> **without** an override being invented for it. Absence is preserved as
> absence — an explicit `false` is a decision, a missing field is not.

---

# Test 4: rollback to the stable snapshot

A different code path: when the remote feed is unreachable, the plugin falls
back to the last known-good snapshot on disk and repopulates from it.

## Block the CTI host

```bash
CTI_IPS=$(getent ahostsv4 api.pre.cloud.wazuh.com | awk '{print $1}' | sort -u)
for ip in $CTI_IPS; do
  sudo iptables -I OUTPUT -d "$ip" -p tcp --dport 443 -j REJECT
done
curl -sS --max-time 5 https://api.pre.cloud.wazuh.com/ ; echo "exit=$?"
```

```
curl: (7) Failed to connect ... after 32 ms: Couldn't connect to server
exit=7
```

> [!NOTE]
> `iptables -j REJECT` is used rather than an `/etc/hosts` entry: the HTTP
> client keeps established keep-alive connections to the already-resolved IP, so
> a DNS-level block has no effect until the node restarts.

## Force the cycle

### REQUEST

```json
POST /.wazuh-cti-consumers/_delete_by_query?refresh=true
{ "query": { "match_all": {} } }
```

### REQUEST

```json
POST /_plugins/_content_manager/update
```

> [!NOTE]
> Tests 2 to 5 share one environment and build on each other, so the numbers
> carry over: the contract checks in test 2 left an override on `iis` too, which
> is why there are now **four** and why the `_version` of the toggled
> integration has advanced well past the `2` seen in test 1.

## Key check: the log shows the rollback and the capture, together

```
[2026-07-29T16:24:54,676][INFO ][c.w.c.c.c.s.AbstractConsumerService] [indexer]
  Remote snapshot failed for consumer [cti:catalog:consumer:ruleset].
  Rolling back from stable snapshot [.../snapshots/ruleset.stable.zip].
[2026-07-29T16:24:54,680][INFO ][c.w.c.c.c.s.SnapshotServiceImpl] [indexer]
  Captured 4 integration enabled override(s) to carry across the snapshot
```

### REQUEST

```json
GET wazuh-threatintel-integrations/_search
{
  "size": 4,
  "query": { "terms": { "document.metadata.title": ["iptables", "aws-fargate", "netflow", "iis"] } },
  "_source": ["document.metadata.title", "document.enabled", "document.user_enabled"]
}
```

### REPLY (relevant fields)

```
iptables      user_enabled=true  enabled=true
aws-fargate   user_enabled=true  enabled=true
netflow       user_enabled=true  enabled=true
iis           user_enabled=true  enabled=true
```

> [!NOTE]
> Verify **before** unblocking: removing the rules triggers a fresh remote sync
> that overwrites the rollback's result, and then you can no longer tell which
> of the two produced it.

## Unblock

```bash
for ip in $CTI_IPS; do
  sudo iptables -D OUTPUT -d "$ip" -p tcp --dport 443 -j REJECT
done
```

---

# Test 5: plan change — CTI reassigns every `document.id`

The strongest check, and the one that justifies matching by title. A plan change
downloads into hidden shadow indices and swaps the alias atomically; CTI serves
a different catalog, so the ids are all new.

## Record the starting ids

```
iptables      4cd06a34-e133-48f3-b32e-50ea6eeae849
aws-fargate   59bf09cf-502e-4550-8990-25e8aa3f635f
netflow       8a5f0987-9d47-4e8d-a4fe-f90a232ac41b
iis           f68e4043-5dc3-48c5-8529-dd7733f49abb
```

## Drop to the free plan

### REQUEST

```json
DELETE /_plugins/_content_manager/subscription
```

### REQUEST

```json
GET /_plugins/_content_manager/subscription
```

### REPLY

```json
{ "message": { "plan": { "name": "Free", "is_public": true }, "is_registered": false }, "status": 200 }
```

### REQUEST

```json
POST /_plugins/_content_manager/update
```

## The consumer moved to a different catalog

### REQUEST

```json
GET .wazuh-cti-consumers/_doc/cti:catalog:consumer:ruleset
```

### REPLY (relevant fields)

```json
{ "_source": {
  "name": "public-ruleset-5",
  "context": "beta3-t1-ruleset-5",
  "resource": ".../contexts/beta3-t1-ruleset-5/consumers/public-ruleset-5",
  "is_public": true,
  "local_offset": 663
}}
```

> [!NOTE]
> It was `beta4-t0-ruleset-5` / `private-ruleset-5` / offset `1239`. Different
> content entirely.

## Key check: new index, new ids, same overrides

### REQUEST

```json
GET wazuh-threatintel-integrations/_search
{
  "size": 4,
  "query": { "terms": { "document.metadata.title": ["iptables", "aws-fargate", "netflow", "iis"] } },
  "_source": ["document.metadata.title", "document.id", "document.enabled", "document.user_enabled"]
}
```

### REPLY (relevant fields)

```
_index = wazuh-threatintel-integrations-b        (was -a)

iptables      id c853b719-db3e-4493-a371-1123fc51bf36   user_enabled=true  enabled=true
aws-fargate   id 1309e2fe-735e-412e-9b7b-1c8fb5b68d0c   user_enabled=true  enabled=true
iis           id 08732726-c878-425a-a218-40ee2ba353e2   user_enabled=true  enabled=true
netflow       id 5e6bf772-82bd-4275-b230-d1d7b66d3db5   user_enabled=true  enabled=true
```

> [!IMPORTANT]
> **Expected, and this is the point of the whole design:** the physical index
> changed (`-a` → `-b`, the shadow swap), **every** `document.id` was reassigned,
> and all four overrides still landed on the right integrations. Keyed by
> `document.id`, not one would have survived.

Restore the Pro subscription afterwards:

### REQUEST

```json
POST /_plugins/_content_manager/subscription
{ "access_token": "<PRO_TOKEN>" }
```

---

# Test 6: create in the `draft` space

## With an explicit `user_enabled: false`

### REQUEST

```json
POST /_plugins/_content_manager/integrations
{
  "resource": {
    "metadata": { "title": "test-enabled-false", "author": "validator",
                  "description": "explicit false", "documentation": "",
                  "references": ["https://wazuh.com"] },
    "category": "cloud-services",
    "user_enabled": false
  }
}
```

### REPLY

```json
{ "message": "ca18bd7e-e1a5-402d-b55a-dc54d9038805", "status": 201 }
```

### REQUEST

```json
GET wazuh-threatintel-integrations/_doc/ca18bd7e-e1a5-402d-b55a-dc54d9038805
```

### REPLY (relevant fields)

```json
{ "_source": {
  "document": { "user_enabled": false, "enabled": false, "mode": "user-managed" },
  "space": { "name": "draft" } } }
```

## With the field omitted — defaults to `true`

### REQUEST

```json
POST /_plugins/_content_manager/integrations
{
  "resource": {
    "metadata": { "title": "test-enabled-default", "author": "validator",
                  "description": "omitted", "documentation": "",
                  "references": ["https://wazuh.com"] },
    "category": "cloud-services"
  }
}
```

### REPLY

```json
{ "message": "3420591a-4d0e-49fa-bcec-8344e3f3677c", "status": 201 }
```

### REPLY of the `GET` (relevant fields)

```json
{ "_source": { "document": { "user_enabled": true, "enabled": true } } }
```

> [!NOTE]
> On create the field is **optional** and defaults to `true`, exactly as
> `enabled` used to. On update it is **required** (test 7).

## With forged `enabled` and `cti_enabled`

### REQUEST

```json
POST /_plugins/_content_manager/integrations
{
  "resource": {
    "metadata": { "title": "test-enabled-forge", "author": "validator",
                  "description": "forged", "documentation": "",
                  "references": ["https://wazuh.com"] },
    "category": "cloud-services",
    "user_enabled": false,
    "enabled": true,
    "cti_enabled": true
  }
}
```

### REPLY

```json
{ "message": "d7bc0ede-ac14-4b3e-9314-02adfde6f817", "status": 201 }
```

### REPLY of the `GET` (relevant fields)

```json
{ "_source": { "document": { "user_enabled": false, "enabled": false } } }
```

> [!NOTE]
> **Expected:** `enabled: false` derived from the override, and no
> `cti_enabled` stored. Both forged fields were dropped.

---

# Test 7: update in the `draft` space

## `user_enabled` is required here

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/ca18bd7e-e1a5-402d-b55a-dc54d9038805
{
  "resource": {
    "metadata": { "title": "test-enabled-false", "author": "validator",
                  "description": "no user_enabled", "documentation": "",
                  "references": ["https://wazuh.com"] },
    "category": "cloud-services",
    "rules": [], "decoders": [], "kvdbs": []
  }
}
```

### REPLY

```json
{ "message": "Missing [user_enabled] field.", "status": 400 }
```

> [!NOTE]
> The field validation runs **before** the engine validation, so the caller gets
> the useful error rather than a confusing `Path '/decoders' not found`.

## A draft update flips the value

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/ca18bd7e-e1a5-402d-b55a-dc54d9038805
{
  "resource": {
    "metadata": { "title": "test-enabled-false", "author": "validator",
                  "description": "now enabled", "documentation": "",
                  "references": ["https://wazuh.com"] },
    "category": "cloud-services",
    "user_enabled": true,
    "rules": [], "decoders": [], "kvdbs": []
  }
}
```

### REPLY

```json
{ "message": "ca18bd7e-e1a5-402d-b55a-dc54d9038805", "status": 200 }
```

### REPLY of the `GET` (relevant fields)

```json
{ "_source": { "document": { "user_enabled": true, "enabled": true } } }
```

---

# Test 8: promotion `draft` → `test` → `custom` carries both fields

## Promote

### REQUEST

```json
GET /_plugins/_content_manager/promote?space=draft
```

### REQUEST

```json
POST /_plugins/_content_manager/promote
{ "space": "draft", "changes": { "PASTE_THE_changes_OBJECT_FROM_THE_GET_ABOVE": true } }
```

### REPLY

```json
{ "message": "Promotion completed successfully.", "status": 200 }
```

Repeat with `space=test` to reach `custom`.

## Key check: the same values in all three spaces

### REQUEST

```json
GET wazuh-threatintel-integrations/_search
{
  "size": 20,
  "query": { "terms": { "document.metadata.title": ["test-enabled-false", "test-enabled-default", "test-enabled-forge"] } },
  "_source": ["document.metadata.title", "document.id", "document.enabled", "document.user_enabled", "space.name"]
}
```

### REPLY (relevant fields)

```
draft   test-enabled-default   user_enabled=true    enabled=true
draft   test-enabled-forge     user_enabled=false   enabled=false
draft   test-enabled-false     user_enabled=true    enabled=true
test    test-enabled-default   user_enabled=true    enabled=true
test    test-enabled-forge     user_enabled=false   enabled=false
test    test-enabled-false     user_enabled=true    enabled=true
custom  test-enabled-default   user_enabled=true    enabled=true
custom  test-enabled-forge     user_enabled=false   enabled=false
custom  test-enabled-false     user_enabled=true    enabled=true
```

> [!NOTE]
> **Expected:** both fields copied unchanged through both promotions, including
> the integration created disabled. The promoted copies keep the same
> `document.id` but get an auto-generated OpenSearch `_id`.

---

# Test 9: an integration in `custom` cannot be updated

### REQUEST

```json
PUT /_plugins/_content_manager/integrations/<OPENSEARCH_ID_OF_THE_CUSTOM_COPY>
{
  "resource": {
    "metadata": { "title": "test-enabled-false", "author": "validator",
                  "description": "edit in custom", "documentation": "",
                  "references": ["https://wazuh.com"] },
    "category": "cloud-services",
    "user_enabled": false,
    "rules": [], "decoders": [], "kvdbs": []
  }
}
```

### REPLY

```json
{ "message": "Invalid space value. Must be one of: [draft, standard].", "status": 400 }
```

> [!NOTE]
> `test` and `custom` have no write endpoint for integrations; documents reach
> them only through promotion. To change the enabled state of an integration
> living in `custom`, edit its `draft` copy and promote the change through the
> chain.

---

# Cleanup

### REQUEST

```json
DELETE /_plugins/_content_manager/integrations/ca18bd7e-e1a5-402d-b55a-dc54d9038805
```

Repeat for the other two draft integrations, then promote once more
(`draft → test → custom`) so the deletions propagate.

</details>

---

# Summary of expected results

| Test | Expectation |
| --- | --- |
| 1. Toggle in the `standard` space | `user_enabled` stored; `enabled` derived from it; no `cti_enabled` invented |
| 2. Request contract | `enabled` alone → **400**; `enabled` alongside `user_enabled` → **ignored**; `cti_enabled` → **discarded**; `protected` → **400** |
| 3. Full resynchronisation | Override **survives**; `_version` and `metadata.modified` prove the document was genuinely rebuilt; an untouched integration still follows CTI with **no** override invented |
| 4. Rollback to the stable snapshot | Log shows the rollback **and** `Captured N override(s)`; all overrides survive |
| 5. Plan change (shadow swap) | Physical index changes, **every** `document.id` is reassigned, and all overrides still land — the proof that matching is by title |
| 6. Create in `draft` | `false` honoured; omitted defaults to `true`; forged `enabled`/`cti_enabled` dropped |
| 7. Update in `draft` | `user_enabled` **required** (unlike create); flipping it re-derives `enabled` |
| 8. Promotion `draft` → `test` → `custom` | Both fields copied unchanged through both hops |
| 9. `PUT` against `custom` | **400** `Invalid space value. Must be one of: [draft, standard].` |

PR #275, opened 2025-02-12
Update deploy-docs.yml

AlexRuiz7:
### Description
This PR fixes a bug on the workflow to deploy the documentation to GH Pages.

Result: https://github.com/wazuh/wazuh-indexer-plugins/actions/runs/13291954727

### Issues Resolved
Closes #254

PR #1465, opened 2026-08-20
Reject promotions that would leave a detector without enabled rules

### LINKED ISSUE #1394: [BUG] `enabled` flag not evaluated on various resource-related operations
## Summary

Across multiple resources and operations (Spaces promotion, Decoders, Rules, and Detectors), the `enabled`/disabled status of an entity does not appear to have any real functional effect. Disabled entities continue to behave as if they were enabled — they can be promoted, set as root, evaluated in Log Test, and picked up by Detectors.

### 1. Disabled spaces can be promoted regardless of integration/decoder/entity changes (1.1)
Spaces can be promoted even without any changes to integrations, decoders, or other entities, and also when such changes exist. It's unclear whether this is intentional behavior or a gap in the promotion validation logic.

### 2. Disabled decoder can be set as root decoder, but promotion behavior is inconsistent (4.2)
- A **disabled** decoder can still be assigned/set as the **root decoder**.
- However, if the root decoder remains disabled, **promotion fails**.
- Separately, it is possible to **disable the decoder that is currently defined as root**, with no validation preventing this.

This creates an inconsistent state: the system allows configuring a disabled decoder as root, but only discovers the problem at promotion time — and doesn't stop you from disabling an already-assigned root decoder in the first place.

### 3. Disabling a rule and promoting still evaluates it in Log Test (6)
When a rule is disabled and then promoted, it is still evaluated during **Log Test**. This suggests the `enabled` flag on a rule has no functional effect on rule evaluation post-promotion.

### 4. Enabled and disabled rules appear identically in Detectors (7)
Both enabled and disabled rules show up the same way when viewed/selected from the Detectors section. There is no visual or functional distinction based on the `enabled` field, reinforcing the concern from Q3 that this field may not be wired up to any actual behavior.

### 5. Detectors do not filter to only enabled rules
Detectors do not correctly filter their rule set to only include enabled rules — the same underlying issue as Q3/Q4. This raises a broader question:

## Impact

- Users may disable a decoder/rule expecting it to be excluded from evaluation or promotion, but it continues to be active — this is misleading and could cause incorrect detections or false confidence that a rule/decoder is inactive.
- The root decoder flow is particularly risky: a disabled decoder can silently be root until promotion fails, and there's no guardrail against disabling the active root decoder.

## References
- https://github.com/wazuh/internal-devel-requests/issues/5464

ferooo23:
## Description

A promotion into `custom` can leave a detector with no enabled rules, and when it does the detector stops without saying so. Its rules are what get compiled into the alerting monitor, so taking the last enabled one away deletes the monitor and the workflow while the detector stays listed reporting `enabled: true`.

Security Analytics does raise the error, but it never reaches the user. Two layers absorb it, each for a good reason: wazuh/wazuh-indexer-security-analytics#275 downgrades it to a warning so an unrelated detector cannot fail a rule update, and Content Manager's SAP sync is best-effort so a transient problem cannot tumble a whole promotion. The promotion returns `200`.

Detecting it afterwards would mean breaking one of those two, so this rejects the promotion up front.

Relates to #1394. Depends on wazuh/wazuh-indexer-security-analytics#275.

## Proposed Changes

**The check**, in three pieces:

- `DetectorRuleGuard` is the decision, for one detector: does this promotion take it from more than zero enabled rules to zero? It reads the resulting state rather than the operation type, so an update that does not touch `enabled` never blocks and a detector already at zero is left alone.
- `DetectorLookupService` gathers the inputs: the enabled detectors referencing the changed rules, and those rules' `enabled` state in `custom` and in the source space. Plain index reads, because content-manager compiles only against `security-analytics-commons`, which exposes no read action.
- `TransportPostPromoteAction.validateNoDetectorLeftEmptyAsync` is the hook. It runs before the gathering phase, so a rejection leaves nothing to roll back, and only for promotions into `custom`; `draft → test` is never blocked.

**What it reports.** Every detector the promotion would empty, each with the rule that empties it, ordered by detector name. One promotion carries the whole diff and can empty several detectors at once, so naming only the first would have the user fixing them one attempt at a time. The message says what to do: *"Disable or delete the detector, or keep one of its rules enabled."*

**What it lets through.** Detectors the user already stopped: a disabled detector produces nothing either way, and blocking rule changes on its account would forbid a legitimate flow. Rule removals do block, and they now reach the check, because the Security Analytics side made rule deletion space-aware: a detector's reference survives until the `custom` copy goes, which is what the promotion does.

### Results and Evidence

Verified live end-to-end with both plugins deployed. Attaching the validation script and step-by-step Dev Tools guide to this comment.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `DetectorRuleGuardTests` : one case per decision branch, from the last rule being disabled to a removal and an unknown rule state counting as disabled.
- `DetectorLookupServiceTests` : both lookups against a mocked client, including the over-match discarded by the intersection check and a missing detectors index tolerated as empty.
- `TransportPostPromoteActionTests` : the rejection message, with several emptied detectors each named with their own rule, surviving detectors left out, and a deterministic listing order.

## Review Checklist

- [ ] Code changes reviewed

### DIFF
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/DetectorLookupService.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/DetectorLookupService.java
@@ -90,11 +90,13 @@ public class DetectorLookupService {
      *
      * <p>Disabled detectors are filtered out: a stopped detector needs no protection.
      *
-     * <p>Best-effort: if the detectors index cannot be read (for example, it does not exist yet), an
-     * empty list is returned so a promotion is never blocked by a lookup problem.
+     * <p>A failed search is propagated to the caller and fails the promotion: it must not be let
+     * through on a guard that could not read its inputs. A cluster with no detectors index yet is not
+     * one of those failures, because the search uses {@code LENIENT_EXPAND_OPEN} and gets an empty
+     * result instead of an error.
      *
      * @param ruleIds the content-manager rule ids being promoted.
-     * @param listener receives the affected detectors.
+     * @param listener receives the affected detectors, or the failure that prevented reading them.
      */
     public void findDetectorsUsingRules(
             Set<String> ruleIds, ActionListener<List<DetectorRules>> listener) {
@@ -138,7 +140,7 @@ public class DetectorLookupService {
                         },
                         e -> {
                             log.warn(Constants.W_LOG_DETECTOR_LOOKUP_FAILED, e.getMessage());
-                            listener.onResponse(Collections.emptyList());
+                            listener.onFailure(e);
                         }));
     }
 
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/DetectorLookupService.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/DetectorLookupService.java
@@ -1,0 +1,241 @@
+/*
+ * Copyright (C) 2026, Wazuh Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.wazuh.contentmanager.cti.catalog.service;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.lucene.search.join.ScoreMode;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.support.IndicesOptions;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.search.SearchHit;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.transport.client.Client;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import com.wazuh.contentmanager.cti.catalog.model.Space;
+import com.wazuh.contentmanager.utils.Constants;
+
+/**
+ * Reads the data needed to decide whether a promotion would leave a detector without enabled rules:
+ * the detectors currently running and the {@code enabled} state of rules in a given space.
+ *
+ * <p>Contains no decision logic; see {@link
+ * com.wazuh.contentmanager.cti.catalog.utils.DetectorRuleGuard}.
+ */
+public class DetectorLookupService {
+
+    private static final Logger log = LogManager.getLogger(DetectorLookupService.class);
+
+    private static final int MAX_RESULTS = 10000;
+
+    /**
+     * Security Analytics detectors index. Read-only: content-manager never writes to it, and cannot
+     * reference the constant from security-analytics because only its {@code commons} module is on
+     * the compile classpath.
+     */
+    private static final String DETECTORS_INDEX = ".opensearch-sap-detectors-config";
+
+    private static final String CUSTOM_RULES_PATH = "detector.inputs.detector_input.custom_rules";
+
+    private final Client client;
+
+    /**
+     * @param client the OpenSearch client.
+     */
+    public DetectorLookupService(Client client) {
+        this.client = client;
+    }
+
+    /**
+     * A detector and the custom rules it references.
+     *
+     * @param id the detector document id.
+     * @param name the detector name, used in user-facing messages.
+     * @param enabled whether the detector is running.
+     * @param ruleIds the content-manager rule ids referenced through {@code custom_rules}.
+     */
+    public record DetectorRules(String id, String name, boolean enabled, List<String> ruleIds) {}
+
+    /**
+     * Returns the enabled detectors that reference any of the given rules, with the custom rules each
+     * one references.
+     *
+     * <p>The nested query narrows the result server-side. Because {@code custom_rules.id} is mapped
+     * as {@code text}, the match is analysed and can return detectors that merely share a token of a
+     * UUID; those are discarded by an exact intersection check. The match never misses a real
+     * reference, so no detector is silently skipped.
+     *
+     * <p>Disabled detectors are filtered out: a stopped detector needs no protection.
+     *
+     * <p>Best-effort: if the detectors index cannot be read (for example, it does not exist yet), an
+     * empty list is returned so a promotion is never blocked by a lookup problem.
+     *
+     * @param ruleIds the content-manager rule ids being promoted.
+     * @param listener receives the affected detectors.
+     */
+    public void findDetectorsUsingRules(
+            Set<String> ruleIds, ActionListener<List<DetectorRules>> listener) {
+        if (ruleIds.isEmpty()) {
+            listener.onResponse(Collections.emptyList());
+            return;
+        }
+
+        BoolQueryBuilder anyRule = QueryBuilders.boolQuery().minimumShouldMatch(1);
+        for (String ruleId : ruleIds) {
+            anyRule.should(QueryBuilders.matchQuery(CUSTOM_RULES_PATH + ".id", ruleId));
+        }
+
+        SearchRequest request =
+                new SearchRequest(DETECTORS_INDEX)
+                        .source(
+                                new SearchSourceBuilder()
+                                        .query(QueryBuilders.nestedQuery(CUSTOM_RULES_PATH, anyRule, ScoreMode.None))
+                                        .fetchSource(
+                                                new String[] {
+                                                    "detector.name", "detector.enabled", CUSTOM_RULES_PATH + ".id"
+                                                },
+                                                null)
+                                        .size(MAX_RESULTS))
+                        .indicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN);
+
+        this.client.search(
+                request,
+                ActionListener.wrap(
+                        response -> {
+                            List<DetectorRules> detectors = new ArrayList<>();
+                            for (SearchHit hit : response.getHits()) {
+                                DetectorRules parsed = parseDetector(hit);
+                                if (parsed != null
+                                        && parsed.enabled()
+                                        && parsed.ruleIds().stream().anyMatch(ruleIds::contains)) {
+                                    detectors.add(parsed);
+                                }
+                            }
+                            listener.onResponse(detectors);
+                        },
+                        e -> {
+                            log.warn(Constants.W_LOG_DETECTOR_LOOKUP_FAILED, e.getMessage());
+                            listener.onResponse(Collections.emptyList());
+                        }));
+    }
+
+    /** Extracts the name, enabled flag and referenced custom rule ids from a detector document. */
+    @SuppressWarnings("unchecked")
+    private static DetectorRules parseDetector(SearchHit hit) {
+        Map<String, Object> source = hit.getSourceAsMap();
+        Object detectorObj = source == null ? null : source.get(Constants.KEY_DETECTOR);
+        if (!(detectorObj instanceof Map)) {
+            return null;
+        }
+        Map<String, Object> detector = (Map<String, Object>) detectorObj;
+
+        boolean enabled = Boolean.TRUE.equals(detector.get(Constants.KEY_ENABLED));
+        Object name = detector.get(Constants.KEY_NAME);
+
+        List<String> ruleIds = new ArrayList<>();
+        Object inputsObj = detector.get("inputs");
+        if (inputsObj instanceof List) {
+            for (Object inputObj : (List<Object>) inputsObj) {
+                if (!(inputObj instanceof Map)) {
+                    continue;
+                }
+                Object detectorInputObj = ((Map<String, Object>) inputObj).get("detector_input");
+                if (!(detectorInputObj instanceof Map)) {
+                    continue;
+                }
+                Object customRulesObj = ((Map<String, Object>) detectorInputObj).get("custom_rules");
+                if (!(customRulesObj instanceof List)) {
+                    continue;
+                }
+                for (Object ruleObj : (List<Object>) customRulesObj) {
+                    if (ruleObj instanceof Map) {
+                        Object id = ((Map<String, Object>) ruleObj).get(Constants.KEY_ID);
+                        if (id != null) {
+                            ruleIds.add(id.toString());
+                        }
+                    }
+                }
+            }
+        }
+
+        return new DetectorRules(
+                hit.getId(), name == null ? hit.getId() : name.toString(), enabled, ruleIds);
+    }
+
+    /**
+     * Returns the {@code enabled} state of the given rules within a space. A rule whose document has
+     * no {@code enabled} field counts as enabled, matching the convention used elsewhere. Rules that
+     * do not exist in the space are simply absent from the result.
+     *
+     * @param ruleIds the content-manager rule ids to look up.
+     * @param space the space to read from.
+     * @param listener receives rule id to enabled state.
+     */
+    public void fetchRuleEnabledStates(
+            Set<String> ruleIds, Space space, ActionListener<Map<String, Boolean>> listener) {
+        if (ruleIds.isEmpty()) {
+            listener.onResponse(Collections.emptyMap());
+            return;
+        }
+
+        SearchRequest request =
+                new SearchRequest(Constants.INDEX_RULES)
+                        .source(
+                                new SearchSourceBuilder()
+                                        .query(
+                                                QueryBuilders.boolQuery()
+                                                        .must(QueryBuilders.termsQuery(Constants.Q_DOCUMENT_ID, ruleIds))
+                                                        .must(
+                                                                QueryBuilders.termQuery(Constants.Q_SPACE_NAME, space.toString())))
+                                        .fetchSource(
+                                                new String[] {Constants.Q_DOCUMENT_ID, Constants.Q_DOCUMENT_ENABLED}, null)
+                                        .size(MAX_RESULTS));
+
+        this.client.search(
+                request,
+                ActionListener.wrap(
+                        response -> {
+                            Map<String, Boolean> states = new HashMap<>();
+                            for (SearchHit hit : response.getHits()) {
+                                Map<String, Object> source = hit.getSourceAsMap();
+                                Object documentObj = source == null ? null : source.get(Constants.KEY_DOCUMENT);
+                                if (!(documentObj instanceof Map)) {
+                                    continue;
+                                }
+                                @SuppressWarnings("unchecked")
+                                Map<String, Object> document = (Map<String, Object>) documentObj;
+                                Object id = document.get(Constants.KEY_ID);
+                                if (id == null) {
+                                    continue;
+                                }
+                                states.put(
+                                        id.toString(), !Boolean.FALSE.equals(document.get(Constants.KEY_ENABLED)));
+                            }
+                            listener.onResponse(states);
+                        },
+                        listener::onFailure));
+    }
+}
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/utils/DetectorRuleGuard.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/utils/DetectorRuleGuard.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2026, Wazuh Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.wazuh.contentmanager.cti.catalog.utils;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Decides whether promoting a set of rule changes would leave a detector with no enabled rules.
+ *
+ * <p>A detector compiles its enabled rules into an alerting monitor. If every rule it references
+ * ends up disabled, the monitor is removed and the detector stops producing findings while still
+ * appearing configured. This guard lets the promotion be rejected before that happens.
+ */
+public final class DetectorRuleGuard {
+
+    private DetectorRuleGuard() {}
+
+    /**
+     * Returns whether the promotion would take a detector from having at least one enabled rule to
+     * having none. A detector that already had none is not considered, since the promotion does not
+     * make its situation worse.
+     *
+     * @param detectorRuleIds the rule ids referenced by the detector.
+     * @param currentEnabled enabled state of those rules in the target space, before the promotion. A
+     *     missing entry counts as disabled.
+     * @param incomingEnabled enabled state carried by the promotion, keyed by rule id. Only contains
+     *     the rules included in the changeset.
+     * @param removedRuleIds rule ids the promotion deletes.
+     * @return {@code true} when the promotion must be rejected for this detector.
+     */
+    public static boolean wouldLeaveDetectorEmpty(
+            List<String> detectorRuleIds,
+            Map<String, Boolean> currentEnabled,
+            Map<String, Boolean> incomingEnabled,
+            Set<String> removedRuleIds) {
+
+        long before =
+                detectorRuleIds.stream().filter(id -> Boolean.TRUE.equals(currentEnabled.get(id))).count();
+        if (before == 0) {
+            return false;
+        }
+
+        long after =
+                detectorRuleIds.stream()
+                        .filter(id -> !removedRuleIds.contains(id))
+                        .filter(
+                                id ->
+                                        incomingEnabled.containsKey(id)
+                                                ? Boolean.TRUE.equals(incomingEnabled.get(id))
+                                                : Boolean.TRUE.equals(currentEnabled.get(id)))
+                        .count();
+
+        return after == 0;
+    }
+}
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/transport/TransportPostPromoteAction.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/transport/TransportPostPromoteAction.java
@@ -39,12 +39,14 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.wazuh.contentmanager.action.MessageStatusResponse;
 import com.wazuh.contentmanager.action.PostPromoteAction;
@@ -1377,12 +1379,6 @@ public class TransportPostPromoteAction
             return;
         }
 
-        // Removals are accounted for, but today they cannot reach this check: deleting a rule
-        // already strips its id from every detector that referenced it, because Security Analytics
-        // matches references by content-manager document id and that id is shared by the copy of
-        // the rule in every space. By the time the removal is promoted, the detector no longer
-        // references the rule and the lookup below returns nothing. The accounting is kept so the
-        // guard still holds if that cascade is ever made space-aware.
         Set<String> changedRuleIds = new HashSet<>();
         Set<String> removedRuleIds = new HashSet<>();
         for (SpaceDiff.OperationItem item : ruleChanges) {
@@ -1415,7 +1411,7 @@ public class TransportPostPromoteAction
                                                             ActionListener.wrap(
                                                                     incomingEnabled ->
                                                                             listener.onResponse(
-                                                                                    firstRejection(
+                                                                                    rejectionMessage(
                                                                                             affected,
                                                                                             currentEnabled,
                                                                                             incomingEnabled,
@@ -1427,32 +1423,86 @@ public class TransportPostPromoteAction
     }
 
     /**
-     * Returns the rejection message for the first detector the promotion would empty, or {@code null}
+     * Returns the rejection message naming every detector the promotion would empty, or {@code null}
      * when none would be.
+     *
+     * <p>A promotion carries the whole space diff, so a single one can empty several detectors, each
+     * through a different rule. All of them are reported together, otherwise the user would fix one
+     * per attempt without knowing how many are left. The listing is ordered by detector name so the
+     * same promotion always produces the same message.
+     *
+     * @param detectors the enabled detectors referencing the promoted rules.
+     * @param currentEnabled enabled state of those detectors' rules in the target space.
+     * @param incomingEnabled enabled state carried by the promotion.
+     * @param removedRuleIds rule ids the promotion deletes.
+     * @return the message to return as {@code 400}, or {@code null} to let the promotion through.
      */
-    private static String firstRejection(
+    static String rejectionMessage(
             List<DetectorLookupService.DetectorRules> detectors,
             Map<String, Boolean> currentEnabled,
             Map<String, Boolean> incomingEnabled,
             Set<String> removedRuleIds) {
+
+        List<EmptiedDetector> emptied = new ArrayList<>();
         for (DetectorLookupService.DetectorRules detector : detectors) {
             if (DetectorRuleGuard.wouldLeaveDetectorEmpty(
                     detector.ruleIds(), currentEnabled, incomingEnabled, removedRuleIds)) {
-                String culprit =
-                        detector.ruleIds().stream()
-                                .filter(
-                                        id ->
-                                                Boolean.TRUE.equals(currentEnabled.get(id))
-                                                        && (removedRuleIds.contains(id)
-                                                                || Boolean.FALSE.equals(incomingEnabled.get(id))))
-                                .findFirst()
-                                .orElse("");
-                return String.format(
-                        Locale.ROOT, Constants.E_400_PROMOTION_EMPTIES_DETECTOR, culprit, detector.name());
+                emptied.add(
+                        new EmptiedDetector(
+                                detector.name(),
+                                culpritRule(detector, currentEnabled, incomingEnabled, removedRuleIds)));
             }
         }
-        return null;
+
+        if (emptied.isEmpty()) {
+            return null;
+        }
+        emptied.sort(Comparator.comparing(EmptiedDetector::detectorName));
+
+        if (emptied.size() == 1) {
+            EmptiedDetector only = emptied.get(0);
+            return String.format(
+                    Locale.ROOT,
+                    Constants.E_400_PROMOTION_EMPTIES_DETECTOR,
+                    only.culpritRule(),
+                    only.detectorName());
+        }
+
+        String listing =
+                emptied.stream()
+                        .map(
+                                item ->
+                                        String.format(
+                                                Locale.ROOT,
+                                                "[%s] by disabling rule [%s]",
+                                                item.detectorName(),
+                                                item.culpritRule()))
+                        .collect(Collectors.joining(", "));
+        return String.format(
+                Locale.ROOT, Constants.E_400_PROMOTION_EMPTIES_DETECTORS, emptied.size(), listing);
     }
+
+    /**
+     * Returns the first rule of the detector that the promotion takes away, which is the one worth
+     * naming to the user.
+     */
+    private static String culpritRule(
+            DetectorLookupService.DetectorRules detector,
+            Map<String, Boolean> currentEnabled,
+            Map<String, Boolean> incomingEnabled,
+            Set<String> removedRuleIds) {
+        return detector.ruleIds().stream()
+                .filter(
+                        id ->
+                                Boolean.TRUE.equals(currentEnabled.get(id))
+                                        && (removedRuleIds.contains(id)
+                                                || Boolean.FALSE.equals(incomingEnabled.get(id))))
+                .findFirst()
+                .orElse("");
+    }
+
+    /** A detector the promotion would empty, with the rule that empties it. */
+    private record EmptiedDetector(String detectorName, String culpritRule) {}
 
     // ── Inner classes ────────────────────────────────────────────────────────
 
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/transport/TransportPostPromoteAction.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/transport/TransportPostPromoteAction.java
@@ -52,8 +52,10 @@ import com.wazuh.contentmanager.action.PostPromoteRequest;
 import com.wazuh.contentmanager.action.ReloadEngineContentAction;
 import com.wazuh.contentmanager.action.ReloadEngineContentRequest;
 import com.wazuh.contentmanager.cti.catalog.model.Space;
+import com.wazuh.contentmanager.cti.catalog.service.DetectorLookupService;
 import com.wazuh.contentmanager.cti.catalog.service.SecurityAnalyticsService;
 import com.wazuh.contentmanager.cti.catalog.service.SpaceService;
+import com.wazuh.contentmanager.cti.catalog.utils.DetectorRuleGuard;
 import com.wazuh.contentmanager.engine.service.EngineService;
 import com.wazuh.contentmanager.rest.model.SpaceDiff;
 import com.wazuh.contentmanager.utils.Constants;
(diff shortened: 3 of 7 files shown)


ferooo23:
# Validation 🟢

> [!IMPORTANT]
> **Prerequisites.** The Security Analytics side must be deployed too
> (wazuh/wazuh-indexer-security-analytics#275). Case 9 only reaches the guard because rule
> deletion is space-aware there now: before that, the `DELETE` stripped the rule id from every
> detector that referenced it and the guard found nothing left to protect.

> [!NOTE]
> These checks can be run automatically with the attached script:
> `IDX=https://localhost:9200 AUTH=admin:admin bash validate-issue-1394-promotion-guard.sh`
>
> `SETTLE=10` raises the wait after each promotion on a loaded cluster.
>
> **Download the script:**  [validate-issue-1394-promotion-guard.sh](https://github.com/user-attachments/files/31302893/validate-issue-1394-promotion-guard.sh)


> [!IMPORTANT]
> **Conventions.** A change only becomes live in the space the detector uses (`custom`) after
> **promotion**: a two-step call where `GET .../promote?space=X` returns a `changes` object that
> you paste into `POST .../promote`, promoting `draft` then `test` to reach `custom`. That call is
> what all nine cases exercise.
>
> A rule's `logsource.product` must equal the parent integration's `metadata.title` (here
> `val1465-int`). The guard does not look at it, but four of the cases read the detector's
> compiled queries, and a rule that does not compile leaves nothing to read.
>
> Two behaviours to know before reading the cases:
>
> - **A rejected promotion does not discard the pending change.** It stays in `draft`/`test`,
>   diverged from `custom`, and rides along on the *next* promotion. Every case that expects a
>   specific starting point sets it explicitly.
> - **A `DELETE` takes a moment to become visible**, and a promotion computes its diff from what
>   it can see. Promoting one second after a delete promotes nothing, returns `200`, and leaves
>   the removal pending, which looks exactly like a guard that failed to fire.

## What we are validating

A promotion into `custom` must not leave a detector without enabled rules, and must not get in
the way of anything else.

| # | Promotion | Expected |
| --- | --- | --- |
| 1 | Disables one of a detector's two enabled rules | allowed |
| 2 | Updates a rule without touching `enabled` | allowed |
| 3 | Disables the last enabled rule, `draft → test` | allowed |
| 4 | Disables the last enabled rule, `test → custom` | **rejected** `400` |
| 5 | state after that rejection | nothing written |
| 6 | Empties two detectors through different rules | **rejected** `400`, naming both |
| 7 | Would empty a detector the user already stopped | allowed |
| 8 | Disables a rule no detector references | allowed |
| 9 | Deletes the last enabled rule, `test → custom` | **rejected** `400` |

One integration, four rules, two detectors:

| Detector | Rules | Used by |
| --- | --- | --- |
| `val1465-detector-a` | `A1`, `A2` | cases 1-5, 9 |
| `val1465-detector-b` | `B1` | cases 6, 7 |
| — | `C1` | case 8 |

Every id and message below comes from one real run.

---

<details>
<summary><b>Step-by-step in Dev Tools</b>: expand to review every request and response</summary>

# Setup

## Create the integration

### REQUEST

```json
POST /_plugins/_content_manager/integrations
{
  "resource": {
    "metadata": { "title": "val1465-int", "author": "validator", "references": ["https://wazuh.com"] },
    "category": "cloud-services",
    "enabled": true
  }
}
```

### REPLY

```json
{ "message": "a7dec36a-aaad-4488-93cf-91a55413f556", "status": 201 }
```

## Create the four rules

All four are created enabled. This is `A1`; `A2`, `B1` and `C1` are the same with their own
title, `sigma_id` and marker.

### REQUEST

```json
POST /_plugins/_content_manager/rules
{
  "integration": "<INTEGRATION_ID>",
  "resource": {
    "metadata": { "title": "val-a1", "author": "validator", "references": ["https://wazuh.com"] },
    "sigma_id": "val-a1",
    "enabled": true,
    "status": "experimental",
    "level": "low",
    "logsource": { "product": "val1465-int", "category": "system" },
    "detection": { "condition": "selection", "selection": { "event.action": ["mark_a1"] } }
  }
}
```

### REPLY (relevant fields)

```
A1 = c53e76e0-556e-489a-959a-50e40fc38aa0
A2 = 294cf87e-1a9d-49a2-ad6b-6b621ec65b8c
B1 = cdabc2e3-0648-4e37-8795-03c2a9ad34a7
C1 = 1726edc4-da59-4805-a95a-6df642c5e6d0
```

Promote `draft → test → custom` so the rules reach the space the detectors read.

## Create the two detectors

### REQUEST

```json
POST /_plugins/_security_analytics/detectors
{
  "name": "val1465-detector-a",
  "detector_type": "docker",
  "enabled": true,
  "schedule": { "period": { "interval": 1, "unit": "MINUTES" } },
  "inputs": [{ "detector_input": {
    "description": "",
    "indices": ["wazuh-events-v5-system-activity"],
    "custom_rules": [{ "id": "<A1_ID>" }, { "id": "<A2_ID>" }],
    "pre_packaged_rules": []
  }}],
  "triggers": []
}
```

### REPLY (relevant fields)

```
detector A = LQrwI6ABdbvoIfH8smBO   queries = [mark_a2 || mark_a1]
detector B = MQrwI6ABdbvoIfH8tWBP   queries = [mark_b1]
```

## The query worth keeping at hand

Every case reads as "which copies of this rule exist, and what does each say about `enabled`".

### REQUEST

```json
POST /wazuh-threatintel-rules/_search
{
  "size": 10,
  "query": { "term": { "document.id": "<RULE_ID>" } },
  "_source": ["space.name", "document.enabled"]
}
```

### REPLY (relevant fields)

```
A1  custom=true  draft=true  test=true
```

---

# Case 1: disabling one of two enabled rules is allowed

`PUT` `A2` with `"enabled": false`, then promote `draft` and `test`.

### REPLY

```json
{ "message": "Promotion completed successfully.", "status": 200 }
```

### REPLY (relevant fields)

```
A2 by space        = custom=false  draft=false  test=false
detector A queries = [event.action: "mark_a1"]
```

> [!NOTE]
> **Expected:** allowed. Two enabled rules to one is not an emptying, and the monitor is rebuilt
> with the survivor.

---

# Case 2: updating a rule without touching `enabled` is allowed

`A1` is now detector A's only enabled rule. `PUT` it with the same `"enabled": true` and
`"level": "medium"`, then promote.

### REPLY (relevant fields)

```
A1 level by space = custom=medium  draft=medium  test=medium
A1 by space       = custom=true    draft=true    test=true
```

> [!IMPORTANT]
> **Expected:** allowed, because the guard reads the resulting state and not the operation. The
> level is read back on purpose: a promotion with nothing to promote also returns `200`, so
> without it "allowed" would prove nothing.

---

# Case 3: disabling the last enabled rule is allowed into `test`

`PUT` `A1` with `"enabled": false`, then promote **`draft` only**.

### REPLY

```json
{ "message": "Promotion completed successfully.", "status": 200 }
```

### REPLY (relevant fields)

```
A1 by space = custom=true  draft=false  test=false
```

> [!NOTE]
> **Expected:** allowed. `draft → test` writes nothing a detector reads, so blocking it would
> forbid staging a change for review.

---

# Case 4: the same promotion into `custom` is rejected

### REQUEST

```json
POST /_plugins/_content_manager/promote
{ "space": "test", "changes": { "PASTE_THE_changes_OBJECT_FROM_THE_GET": true } }
```

### REPLY

```json
{
  "error": { "reason": "Rule [c53e76e0-556e-489a-959a-50e40fc38aa0] cannot be promoted because it would leave detector [val1465-detector-a-1787309365] without enabled rules. Disable or delete the detector, or keep one of its rules enabled." },
  "status": 400
}
```

> [!IMPORTANT]
> **Expected:** `400`, naming both the rule and the detector.

---

# Case 5: the rejection wrote nothing

### REQUEST

```json
GET /.opensearch-sap-detectors-config/_doc/<DETECTOR_A_ID>
```

### REPLY (relevant fields)

```json
{ "_source": { "detector": {
  "enabled": true,
  "monitor_id": ["KgrwI6ABdbvoIfH8r2Cs"],
  "workflow_ids": ["LArwI6ABdbvoIfH8smAc"]
} } }
```

### REPLY (relevant fields)

```
A1 by space        = custom=true  draft=false  test=false
detector A queries = [event.action: "mark_a1"]
```

> [!IMPORTANT]
> **Expected:** monitor **and** workflow both alive, and `A1` still enabled in `custom`. The
> damage this guard prevents is losing both, so both are checked.

---

# Case 6: one promotion emptying two detectors names both

`A1` is re-enabled and promoted first. Then `PUT` **both** `A1` and `B1` with
`"enabled": false` and promote: one promotion, two detectors, two different rules.

### REPLY

```json
{
  "error": { "reason": "This promotion would leave 2 detectors without enabled rules: [val1465-detector-a-1787309365] by disabling rule [c53e76e0-556e-489a-959a-50e40fc38aa0], [val1465-detector-b-1787309365] by disabling rule [cdabc2e3-0648-4e37-8795-03c2a9ad34a7]. Disable or delete them, or keep one of their rules enabled." },
  "status": 400
}
```

### REPLY (relevant fields)

```
A1 by space = custom=true  draft=false  test=false
B1 by space = custom=true  draft=false  test=false
```

> [!IMPORTANT]
> **Expected:** one `400` listing **both** detectors, each with the rule that empties it, `a`
> before `b`. The order is checked by comparing where each name falls in the message, not by eye.

---

# Case 7: emptying a detector the user already stopped is allowed

Both rules are re-enabled and promoted first. Then turn detector B off:

### REQUEST

```json
PUT /_plugins/_security_analytics/detectors/<DETECTOR_B_ID>
```

with the detector body and `"enabled": false`.

> [!NOTE]
> The API replaces the whole object, so the natural move is to `GET` the detector and send it
> back with the flag flipped. Drop `last_update_time` and `enabled_time` from that body: `GET`
> serialises them as ISO-8601 strings while the parser expects epoch millis, so echoing them
> verbatim returns `illegal_argument_exception`. The server sets them anyway.

Now `PUT` `B1` with `"enabled": false` and promote.

### REPLY

```json
{ "message": "Promotion completed successfully.", "status": 200 }
```

### REPLY (relevant fields)

```
B1 by space        = custom=false  draft=false  test=false
detector A queries = [event.action: "mark_a1"]
```

> [!IMPORTANT]
> **Expected:** allowed. A stopped detector produces nothing either way. Detector A, which is
> running, is left alone by the same promotion.

---

# Case 8: disabling a rule no detector references is allowed

`PUT` `C1` with `"enabled": false` and promote.

### REPLY (relevant fields)

```
C1 by space = custom=false  draft=false  test=false
```

> [!NOTE]
> **Expected:** allowed. Nothing references it, so there is nothing to empty.

---

# Case 9: deleting the last enabled rule is rejected too

`A1` is detector A's only enabled rule at this point, `A2` having been disabled in case 1.

### REQUEST

```json
DELETE /_plugins/_content_manager/rules/<A1_ID>
```

### REPLY (relevant fields)

```
A1 by space = custom=true  test=true          <- the draft copy is gone
```

Promote `draft → test`:

### REPLY (relevant fields)

```
{ "message": "Promotion completed successfully.", "status": 200 }
A1 by space = custom=true                     <- the removal reached the test copy
```

Promote `test → custom`:

### REPLY

```json
{
  "error": { "reason": "Rule [c53e76e0-556e-489a-959a-50e40fc38aa0] cannot be promoted because it would leave detector [val1465-detector-a-1787309365] without enabled rules. Disable or delete the detector, or keep one of its rules enabled." },
  "status": 400
}
```

### REPLY (relevant fields)

```
A1 by space        = custom=true
detector A queries = [event.action: "mark_a1"]
```

> [!IMPORTANT]
> **Expected:** `400`. A removal is treated like a disable, because the outcome for the detector
> is the same. Each promotion is checked before moving on, for the reason in Conventions.

---

# Cleanup

Delete the detectors **first**. While they still reference the rules, promoting the removals is
precisely what the guard rejects, which would strand the rules and leave the integration
undeletable: its log type cannot be deleted while active rules exist.

```
DELETE /_plugins/_security_analytics/detectors/<DETECTOR_A_ID>
DELETE /_plugins/_security_analytics/detectors/<DETECTOR_B_ID>
DELETE /_plugins/_content_manager/rules/<A1_ID>   (and A2, B1, C1)
```

Promote `draft → test → custom`, then drop the integration and promote once more.

</details>

---

# Summary of results

| # | Promotion | Result |
| --- | --- | --- |
| 1 | Disables one of two enabled rules | allowed, monitor rebuilt with the survivor |
| 2 | Updates a rule without touching `enabled` | allowed, and the update verified in `custom` |
| 3 | Disables the last enabled rule, `draft → test` | allowed, `custom` untouched |
| 4 | Disables the last enabled rule, `test → custom` | rejected `400`, naming rule and detector |
| 5 | State after the rejection | monitor, workflow and `custom` copy all intact |
| 6 | Empties two detectors through different rules | rejected `400`, naming both, ordered by name |
| 7 | Would empty a stopped detector | allowed, the running detector untouched |
| 8 | Disables a rule no detector references | allowed |
| 9 | Deletes the last enabled rule, `test → custom` | rejected `400`, the removal counts like a disable |

felipegonzalezmv:
# Validation evidence — PR #1465

**Reject promotions that would leave a detector without enabled rules**
`1394-block-promotion-emptying-detector` → `5.0.0` · run on 2026-08-21

Test list and assertions taken from the PR's own validation comment,
[wazuh-indexer-plugins#1465 (comment 5369051855)](https://github.com/wazuh/wazuh-indexer-plugins/pull/1465#issuecomment-5369051855),
executed with its attached script `validate-issue-1394-promotion-guard.sh`
(downloaded from the comment) against a clean local cluster. Everything below was
done over the REST API with `curl`; no Dev Tools.

**Result: 9 / 9 cases green — 28 assertions passed, 0 failed.**

| # | Promotion | Expected | Result |
|---|---|---|---|
| 1 | Disables one of a detector's two enabled rules | allowed | 🟢 |
| 2 | Updates a rule without touching `enabled` | allowed | 🟢 |
| 3 | Disables the last enabled rule, `draft → test` | allowed | 🟢 |
| 4 | Disables the last enabled rule, `test → custom` | rejected `400` | 🟢 |
| 5 | State after that rejection | nothing written | 🟢 |
| 6 | Empties two detectors through different rules | rejected `400`, naming both | 🟢 |
| 7 | Would empty a detector the user already stopped | allowed | 🟢 |
| 8 | Disables a rule no detector references | allowed | 🟢 |
| 9 | Deletes the last enabled rule, `test → custom` | rejected `400` | 🟢 |

Every request and reply of the run, verbatim and in order:
[`run-2026-08-21-curl-trace.txt`](run-2026-08-21-curl-trace.txt).
Script output: [`run-2026-08-21-raw.txt`](run-2026-08-21-raw.txt).

---

## 1. Build

`./gradlew :wazuh-indexer-content-manager:build` — full `check`, nothing skipped
(`test`, `spotlessCheck`, `forbiddenApis`, `testingConventions`, `integTest` all ran).

```
BUILD SUCCESSFUL in 3m 45s
```

89 suites · **600 tests · 0 failures · 0 errors**. The three suites the PR adds:

| Suite | Tests | Failures |
|---|---|---|
| `DetectorRuleGuardTests` | 7 | 0 |
| `DetectorLookupServiceTests` | 7 | 0 |
| `TransportPostPromoteActionTests` | 6 | 0 |

```
DetectorRuleGuardTests            testAllowsWhenOtherRulesStayEnabled, testAllowsSwapWithinSamePromotion,
                                  testBlocksWhenLastEnabledRuleIsRemoved, testBlocksWhenLastEnabledRuleIsDisabled,
                                  testTreatsUnknownRuleStateAsDisabled, testAllowsWhenDetectorWasAlreadyEmpty,
                                  testAllowsUpdateThatKeepsRuleEnabled
DetectorLookupServiceTests        testFindDetectorsUsingRulesReturnsRuleIds, testFetchRuleEnabledStates,
                                  testFindDetectorsUsingRulesToleratesMissingIndex, testFindDetectorsUsingRulesSkipsDisabledOnes,
                                  testFindDetectorsUsingRulesWithNoIds, testFindDetectorsUsingRulesDiscardsOverMatches,
                                  testFetchRuleEnabledStatesWithNoIds
TransportPostPromoteActionTests   testNoRejectionWhenEveryDetectorKeepsARule, testSingleDetectorIsNamedWithItsRule,
                                  testRemovedRuleIsReportedAsTheCulprit, testEveryEmptiedDetectorIsReportedWithItsOwnRule,
                                  testListingIsDeterministic, testSurvivingDetectorsAreNotReported
```

The other repos were built too, since the guard cannot be exercised without them:
`wazuh-indexer-common-utils`, `wazuh-indexer-security-analytics`,
`wazuh-indexer-alerting`, `wazuh-indexer-setup`. All `BUILD SUCCESSFUL`.

## 2. Environment

Docker environment torn down and rebuilt from scratch: `docker compose down -v`
(data volume dropped), `wazuh-indexer-dev:local` and `wazuh-dashboard-dev:local`
deleted, image rebuilt `--no-cache` from freshly compiled ZIPs. Base images
(`wazuh-indexer-pkg:0806`) kept — they come from a `.deb` build that is not
cheaply reproducible.

Host port 9200 is held by a native systemd `wazuh-indexer`, so the container was
mapped to **9201** instead of stopping the host service. Every command below uses:

```bash
IDX=https://localhost:9201
AUTH=admin:admin
```

```bash
curl -sk -u admin:admin $IDX/_cluster/health
```
```json
{"cluster_name":"wazuh-cluster","status":"green","number_of_nodes":1,
 "active_shards":57,"active_shards_percent_as_number":100.0}
```

```bash
curl -sk -u admin:admin "$IDX/_cat/plugins?h=component,version"
```
```
wazuh-indexer-alerting            5.0.0.0
wazuh-indexer-content-manager     5.0.0.0     ← PR #1465
wazuh-indexer-notifications       5.0.0.0
wazuh-indexer-reports-scheduler   5.0.0.0
wazuh-indexer-security-analytics  5.0.0.0     ← PR #275 (required dependency)
wazuh-indexer-setup               5.0.0.0
```

Branches under test — every repo on its default `5.0.0` except the two PRs:

| Repo | Branch |
|---|---|
| `wazuh-indexer-plugins` | `1394-block-promotion-emptying-detector` (PR #1465) |
| `wazuh-indexer-security-analytics` | `fix/1394-detectors-filter-disabled-rules` (PR #275) |

Setup plugin finished before the run started:

```bash
curl -sk -u admin:admin $IDX/.wazuh-setup-status/_doc/setup-status
```
```json
{"status":"ready","timestamp":"2026-08-21T14:36:06.358941367Z"}
```

### 2.1 One environment precondition the comment does not mention

`TransportIndexDetectorAction.importRules()` searches
`.opensearch-sap-pre-packaged-rules-config` unconditionally, even when a detector's
`pre_packaged_rules` is empty:

```java
SearchRequest searchRequest = new SearchRequest(Rule.PRE_PACKAGED_RULES_INDEX)...
```

Wazuh creates that index only under the test-only `default_rules.enabled=true`
switch and ships no bundled sigma rules, so on a cluster with no CTI content it
never exists and **every** detector creation returns:

```json
{"error":{"type":"index_not_found_exception",
          "reason":"no such index [.opensearch-sap-pre-packaged-rules-config]"},"status":404}
```

Created empty with SAP's own `mappings/rules.json` plus `index.hidden` — exactly
what `RuleIndices.initRuleIndex()` applies — by
[`scripts/prep-env-1465.sh`](../../scripts/prep-env-1465.sh):

```bash
curl -sk -u admin:admin -X PUT $IDX/.opensearch-sap-pre-packaged-rules-config \
  -H 'Content-Type: application/json' \
  -d "$(python3 -c "
import json
m=json.load(open('\$SAP_REPO/src/main/resources/mappings/rules.json'))
print(json.dumps({'settings':{'index.hidden':True},'mappings':m}))")"
```
```json
{"acknowledged":true,"shards_acknowledged":true,
 "index":".opensearch-sap-pre-packaged-rules-config"}
```

This is a pre-existing `5.0.0` gap in a CTI-less environment, not a defect in
PR #1465.

### 2.2 No Pro subscription and no CTI content were needed

The nine cases create every rule they use through Content Manager, and promotion
`draft → test → custom` works on a `Free`, unregistered cluster:

```bash
curl -sk -u admin:admin $IDX/_plugins/_content_manager/subscription
```
```json
{"message":{"plan":{"name":"Free","is_public":true},"is_registered":false},"status":200}
```

```bash
curl -sk -u admin:admin "$IDX/wazuh-threatintel-rules*/_count"
```
```json
{"count":0,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0}}
```

### 2.3 Two script deviations, assertions untouched

Run with [`validate-issue-1394-promotion-guard.local.sh`](../../scripts/validate-issue-1394-promotion-guard.local.sh),
a copy of the attached script with exactly this diff:

```diff
-DETECTOR_TYPE="${DETECTOR_TYPE:-docker}"
 TITLE="val1465-int-${SUFFIX}"
+# LOCAL: no CTI content, so the integration's own log type is the only one.
+DETECTOR_TYPE="${DETECTOR_TYPE:-$TITLE}"

-[ "$RCOUNT" -gt 0 ] 2>/dev/null || die "no CTI rules loaded yet — wait for the initial consumer sync"
+[ "$RCOUNT" -gt 0 ] 2>/dev/null || info "no CTI rules loaded (not required by these cases)"
```

- `detector_type` must name an existing log type. With no CTI content the only one
  on the cluster is the log type the integration itself creates, named after the
  integration title. This is invisible to the guard: custom rules are resolved by
  `rule.document.id` + `rule.space == "custom"`, never by `detector_type`
  (`TransportIndexDetectorAction.importCustomRules`, line 2520).
- The CTI preflight is inherited boilerplate; it gates on content none of the nine
  cases consume. Downgraded from `die` to a warning.

Every assertion, threshold and expected message is unmodified. The run was made
with a `--trace`-style wrapper
([`…-promotion-guard.traced.sh`](../../scripts/validate-issue-1394-promotion-guard.traced.sh))
that logs each request as a reproducible `curl` one-liner with its reply; it wraps
the script's own `CURL` array and changes no logic.

---

## 3. Setup — one integration, four rules, two detectors

```bash
IDX=https://localhost:9201 AUTH=admin:admin SETTLE=10 \
  bash validate-issue-1394-promotion-guard.local.sh
```

### Create the integration

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/integrations \
  -d '{"resource":{"metadata":{"title":"val1465-int-1787322981","author":"validator",
       "references":["https://wazuh.com"]},"category":"cloud-services","enabled":true}}'
```
```json
{"message":"235a7b5f-227b-4d7b-a460-6643db49ee56","status":201}
```

### Create the four rules

All four enabled. `A1` shown; `A2`, `B1`, `C1` are identical but for title,
`sigma_id` and marker.

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/rules \
  -d '{"integration":"235a7b5f-227b-4d7b-a460-6643db49ee56","resource":{
       "metadata":{"title":"val-a1-1787322981","author":"validator","references":["https://wazuh.com"]},
       "sigma_id":"val-a1-1787322981","enabled":true,"status":"experimental","level":"low",
       "logsource":{"product":"val1465-int-1787322981","category":"system"},
       "detection":{"condition":"selection","selection":{"event.action":["mark_a1_1787322981"]}}}}'
```
```json
{"message":"127b0f15-98fe-429d-9cfa-7c7e71e6a427","status":201}
```

| Rule | `document.id` | Marker | Detector |
|---|---|---|---|
| `A1` | `127b0f15-98fe-429d-9cfa-7c7e71e6a427` | `mark_a1_1787322981` | A |
| `A2` | `49b61577-76a2-446f-8117-0d3ca27d6ca1` | `mark_a2_1787322981` | A |
| `B1` | `ebc8dbc7-a75b-460c-b916-9630fa915cba` | `mark_b1_1787322981` | B |
| `C1` | `e2845de0-3719-400f-92c9-21881f3ef7b5` | `mark_c1_1787322981` | — |

### Promote `draft → test → custom`

Two steps per hop: `GET` returns the `changes` object, `POST` applies it.

```bash
curl -sk -u admin:admin "$IDX/_plugins/_content_manager/promote?space=draft"
```
```json
{"changes":{"kvdbs":[],"rules":[
  {"operation":"add","id":"49b61577-76a2-446f-8117-0d3ca27d6ca1"},
  {"operation":"add","id":"e2845de0-3719-400f-92c9-21881f3ef7b5"},
  {"operation":"add","id":"ebc8dbc7-a75b-460c-b916-9630fa915cba"},
  {"operation":"add","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}],
  "decoders":[],"filters":[],
  "integrations":[{"operation":"add","id":"235a7b5f-227b-4d7b-a460-6643db49ee56"}],"policy":[…]}}
```

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/promote \
  -d '{"space":"draft","changes":{ … the object above … }}'
```
```json
{"message":"Promotion completed successfully.","status":200}
```

Same pair with `space=test` to reach `custom`:

```bash
curl -sk -u admin:admin "$IDX/_plugins/_content_manager/promote?space=test"
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/promote -d '{"space":"test","changes":{ … }}'
```
```json
{"message":"Promotion completed successfully.","status":200}
```

### Create the two detectors

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_security_analytics/detectors -d '{
    "name":"val1465-detector-a-1787322981","detector_type":"val1465-int-1787322981","enabled":true,
    "schedule":{"period":{"interval":1,"unit":"MINUTES"}},
    "inputs":[{"detector_input":{"description":"","indices":["wazuh-events-v5-system-activity"],
      "custom_rules":[{"id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"},
                      {"id":"49b61577-76a2-446f-8117-0d3ca27d6ca1"}],"pre_packaged_rules":[]}}],
    "triggers":[]}'
```
```json
{"_id":"a8fAJKAB24A6bbTjjFrv","_version":1,"detector":{"name":"val1465-detector-a-1787322981", … }}
```

Detector B the same, referencing `B1` only → `_id` `b8fAJKAB24A6bbTjjlqO`.

### The two read helpers used throughout

Monitor id of a detector:

```bash
curl -sk -u admin:admin $IDX/.opensearch-sap-detectors-config/_doc/a8fAJKAB24A6bbTjjFrv \
  | jq -r '._source.detector.monitor_id[0]'
# → aMfAJKAB24A6bbTji1pr
```

Compiled queries of that monitor — this is the "what is actually running" check:

```bash
curl -sk -u admin:admin $IDX/_plugins/_alerting/monitors/aMfAJKAB24A6bbTji1pr \
  | jq -r '[.monitor.inputs[].doc_level_input.queries[]?.query] | join(" || ")'
# → event.action: "mark_a2_1787322981" || event.action: "mark_a1_1787322981"
```

A rule's `enabled` state in every space that still holds a copy:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  $IDX/wazuh-threatintel-rules/_search \
  -d '{"size":10,"query":{"term":{"document.id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}},
       "_source":["space.name","document.enabled"]}' \
  | jq -r '[.hits.hits[]._source | "\(.space.name)=\(.document.enabled)"] | sort | join(" ")'
# → custom=true draft=true test=true
```

Starting state:

```
detector A = a8fAJKAB24A6bbTjjFrv  -> A1, A2
detector B = b8fAJKAB24A6bbTjjlqO  -> B1
A queries = [event.action: "mark_a2_…" || event.action: "mark_a1_…"]
B queries = [event.action: "mark_b1_…"]
```

---

## 4. The nine cases

### 🟢 Case 1 — disabling one of two enabled rules is allowed

Disable `A2`, leaving `A1` enabled:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X PUT $IDX/_plugins/_content_manager/rules/49b61577-76a2-446f-8117-0d3ca27d6ca1 \
  -d '{"resource":{
       "metadata":{"title":"val-a2-1787322981","author":"validator","references":["https://wazuh.com"]},
       "sigma_id":"val-a2-1787322981","enabled":false,"status":"experimental","level":"low",
       "logsource":{"product":"val1465-int-1787322981","category":"system"},
       "detection":{"condition":"selection","selection":{"event.action":["mark_a2_1787322981"]}}}}'
```
```json
{"message":"49b61577-76a2-446f-8117-0d3ca27d6ca1","status":200}
```

Promote `draft → test → custom`:

```json
{"message":"Promotion completed successfully.","status":200}
```

```
PASS  promotion into custom accepted: detector A keeps A1
PASS  detector A rebuilt with the surviving rule only
      queries = [event.action: "mark_a1_1787322981"]
PASS  A2 disabled in every space
      by space = [custom=false draft=false test=false]
```

### 🟢 Case 2 — updating a rule without touching `enabled` is allowed

`A1` updated with `level` `low → medium`, `enabled` left `true`:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X PUT $IDX/_plugins/_content_manager/rules/127b0f15-98fe-429d-9cfa-7c7e71e6a427 \
  -d '{"resource":{ … "enabled":true, "level":"medium", … }}'
```
```json
{"message":"127b0f15-98fe-429d-9cfa-7c7e71e6a427","status":200}
```

Promote, then read the field back per space to prove the update actually landed
rather than being silently dropped:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  $IDX/wazuh-threatintel-rules/_search \
  -d '{"size":10,"query":{"term":{"document.id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}},
       "_source":["space.name","document.level"]}' \
  | jq -r '[.hits.hits[]._source | "\(.space.name)=\(.document.level)"] | sort | join(" ")'
```

```
PASS  promotion into custom accepted: the update does not touch enabled
PASS  the update actually landed in custom
      A1 level by space = [custom=medium draft=medium test=medium]
PASS  A1 still enabled everywhere
      by space = [custom=true draft=true test=true]
```

### 🟢 Case 3 — disabling the last enabled rule is allowed into `test`

`A1` is now detector A's only enabled rule. Disable it:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X PUT $IDX/_plugins/_content_manager/rules/127b0f15-98fe-429d-9cfa-7c7e71e6a427 \
  -d '{"resource":{ … "enabled":false, "level":"medium", … }}'
```
```json
{"message":"127b0f15-98fe-429d-9cfa-7c7e71e6a427","status":200}
```

Promote **`draft → test` only**:

```bash
curl -sk -u admin:admin "$IDX/_plugins/_content_manager/promote?space=draft"
```
```json
{"changes":{"kvdbs":[],"rules":[{"operation":"update","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}],
 "decoders":[],"filters":[],"integrations":[],"policy":[]}}
```
```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/promote \
  -d '{"space":"draft","changes":{"kvdbs":[],"rules":[{"operation":"update","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}],"decoders":[],"filters":[],"integrations":[],"policy":[]}}'
```
```json
{"message":"Promotion completed successfully.","status":200}
```

```
PASS  promotion draft -> test accepted
PASS  the disable reached test, custom untouched
      by space = [custom=true draft=false test=false]
```

`draft → test` is never guarded — only `test → custom` is.

### 🟢 Case 4 — the same promotion into `custom` is rejected

Same pending change, next hop:

```bash
curl -sk -u admin:admin "$IDX/_plugins/_content_manager/promote?space=test"
```
```json
{"changes":{"kvdbs":[],"rules":[{"operation":"update","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}],
 "decoders":[],"filters":[],"integrations":[],"policy":[]}}
```
```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/promote \
  -d '{"space":"test","changes":{"kvdbs":[],"rules":[{"operation":"update","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}],"decoders":[],"filters":[],"integrations":[],"policy":[]}}'
```
```json
{"message":"Rule [127b0f15-98fe-429d-9cfa-7c7e71e6a427] cannot be promoted because it would leave detector [val1465-detector-a-1787322981] without enabled rules. Disable or delete the detector, or keep one of its rules enabled.","status":400}
```

```
PASS  promotion into custom rejected, naming the rule and the detector
```

`400`, naming both the culprit rule and the detector, with the remediation the PR
description promises.

### 🟢 Case 5 — the rejection wrote nothing

Three reads after the settle window — asserted once, not on a retry loop, so a
late rebuild could not hide.

```bash
# rule still enabled in custom
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  $IDX/wazuh-threatintel-rules/_search \
  -d '{"size":10,"query":{"term":{"document.id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}},
       "_source":["space.name","document.enabled"]}'

# detector still compiles it
curl -sk -u admin:admin $IDX/_plugins/_alerting/monitors/aMfAJKAB24A6bbTji1pr \
  | jq -r '[.monitor.inputs[].doc_level_input.queries[]?.query] | join(" || ")'

# monitor and workflow both still referenced
curl -sk -u admin:admin $IDX/.opensearch-sap-detectors-config/_doc/a8fAJKAB24A6bbTjjFrv \
  | jq -r '._source.detector | "monitor=\(.monitor_id[0]) workflow=\(.workflow_ids[0])"'
```

```
PASS  A1 still enabled in custom
      by space = [custom=true draft=false test=false]
PASS  detector A still compiles A1
      queries = [event.action: "mark_a1_1787322981"]
PASS  monitor and workflow both intact
      monitor = aMfAJKAB24A6bbTji1pr   workflow = asfAJKAB24A6bbTjjFqd
```

Confirms the hook's placement claim: the check runs before the gathering phase, so
a rejection leaves nothing to roll back.

### 🟢 Case 6 — one promotion emptying two detectors names both

`A1` restored to enabled first, then both `A1` and `B1` disabled in the *same*
pending changeset:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X PUT $IDX/_plugins/_content_manager/rules/127b0f15-98fe-429d-9cfa-7c7e71e6a427 \
  -d '{"resource":{ … "enabled":false … }}'
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X PUT $IDX/_plugins/_content_manager/rules/ebc8dbc7-a75b-460c-b916-9630fa915cba \
  -d '{"resource":{ … "enabled":false … }}'
```

Promote `draft → test` (accepted), then `test → custom`:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/promote \
  -d '{"space":"test","changes":{"kvdbs":[],"rules":[
        {"operation":"update","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"},
        {"operation":"update","id":"ebc8dbc7-a75b-460c-b916-9630fa915cba"}],
       "decoders":[],"filters":[],"integrations":[],"policy":[]}}'
```
```json
{"message":"This promotion would leave 2 detectors without enabled rules: [val1465-detector-a-1787322981] by disabling rule [127b0f15-98fe-429d-9cfa-7c7e71e6a427], [val1465-detector-b-1787322981] by disabling rule [ebc8dbc7-a75b-460c-b916-9630fa915cba]. Disable or delete them, or keep one of their rules enabled.","status":400}
```

```
PASS  promotion rejected, naming both detectors and both rules
PASS  the listing is ordered by detector name (a before b)
PASS  A1 untouched in custom      by space = [custom=true draft=false test=false]
PASS  B1 untouched in custom      by space = [custom=true draft=false test=false]
```

Both detectors reported in one response, each with its own culprit rule, ordered
by name — the point of `rejectionMessage` sorting rather than reporting the first
hit.

### 🟢 Case 7 — emptying a detector the user already stopped is allowed

Both rules restored, then detector B turned off. The API replaces the whole
object, so it is read back and patched — `last_update_time` and `enabled_time` are
dropped because `GET` serialises them as ISO-8601 while the parser expects epoch
millis:

```bash
BODY=$(curl -sk -u admin:admin $IDX/_plugins/_security_analytics/detectors/b8fAJKAB24A6bbTjjlqO \
  | jq -c '.detector | .enabled = false | del(.last_update_time, .enabled_time)')

curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X PUT $IDX/_plugins/_security_analytics/detectors/b8fAJKAB24A6bbTjjlqO -d "$BODY"
```

Disable `B1` — the stopped detector's only rule — and promote to `custom`:

```json
{"message":"Promotion completed successfully.","status":200}
```

```
PASS  detector B turned off
PASS  promotion into custom accepted for the stopped detector
PASS  B1 disabled in every space   by space = [custom=false draft=false test=false]
PASS  detector A untouched by it   queries = [event.action: "mark_a1_1787322981"]
```

### 🟢 Case 8 — disabling a rule no detector references is allowed

`C1` is referenced by no detector:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X PUT $IDX/_plugins/_content_manager/rules/e2845de0-3719-400f-92c9-21881f3ef7b5 \
  -d '{"resource":{ … "enabled":false … }}'
```

Promote to `custom`:

```json
{"message":"Promotion completed successfully.","status":200}
```

```
PASS  promotion into custom accepted, no detector involved
PASS  C1 disabled in every space   by space = [custom=false draft=false test=false]
```

### 🟢 Case 9 — deleting the last enabled rule is rejected too

`A1` restored to enabled in all three spaces, then deleted outright:

```bash
curl -sk -u admin:admin -X DELETE \
  $IDX/_plugins/_content_manager/rules/127b0f15-98fe-429d-9cfa-7c7e71e6a427
```

The `DELETE` removes the `draft` copy only; the script waits for that state before
promoting, because a promotion computes its diff from what it can see and
promoting too early would promote nothing and return a misleading `200`:

```
copies left = [custom=true test=true]
```

Promote `draft → test` — note the operation is now `remove`:

```bash
curl -sk -u admin:admin "$IDX/_plugins/_content_manager/promote?space=draft"
```
```json
{"changes":{"kvdbs":[],"rules":[{"operation":"remove","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}],
 "decoders":[],"filters":[],
 "integrations":[{"operation":"update","id":"235a7b5f-227b-4d7b-a460-6643db49ee56"}],"policy":[]}}
```
```json
{"message":"Promotion completed successfully.","status":200}
```
```
copies left = [custom=true]
```

Promote `test → custom`:

```bash
curl -sk -u admin:admin -H 'Content-Type: application/json' \
  -X POST $IDX/_plugins/_content_manager/promote \
  -d '{"space":"test","changes":{"kvdbs":[],"rules":[{"operation":"remove","id":"127b0f15-98fe-429d-9cfa-7c7e71e6a427"}],"decoders":[],"filters":[],"integrations":[{"operation":"update","id":"235a7b5f-227b-4d7b-a460-6643db49ee56"}],"policy":[]}}'
```
```json
{"message":"Rule [127b0f15-98fe-429d-9cfa-7c7e71e6a427] cannot be promoted because it would leave detector [val1465-detector-a-1787322981] without enabled rules. Disable or delete the detector, or keep one of its rules enabled.","status":400}
```

```
PASS  the DELETE removed the draft copy and left the others alone
PASS  promotion draft -> test accepted
PASS  the removal reached the test copy
PASS  promotion into custom rejected, the removal counts like a disable
PASS  the custom copy survived the rejection    by space = [custom=true]
PASS  detector A still compiles A1              queries = [event.action: "mark_a1_1787322981"]
```

A `REMOVE` is treated like a disable, and this only reaches the guard because SAP
#275 made rule deletion space-aware — the dependency the PR declares, exercised.

---

## 5. Cleanup and summary

Detectors are deleted **first**: while they still reference the rules, promoting
the removals is exactly what the guard rejects, which would strand the rules and
leave the integration undeletable.

```bash
curl -sk -u admin:admin -X DELETE $IDX/_plugins/_security_analytics/detectors/a8fAJKAB24A6bbTjjFrv
curl -sk -u admin:admin -X DELETE $IDX/_plugins/_security_analytics/detectors/b8fAJKAB24A6bbTjjlqO
curl -sk -u admin:admin -X DELETE $IDX/_plugins/_content_manager/rules/127b0f15-98fe-429d-9cfa-7c7e71e6a427
curl -sk -u admin:admin -X DELETE $IDX/_plugins/_content_manager/rules/49b61577-76a2-446f-8117-0d3ca27d6ca1
curl -sk -u admin:admin -X DELETE $IDX/_plugins/_content_manager/rules/ebc8dbc7-a75b-460c-b916-9630fa915cba
curl -sk -u admin:admin -X DELETE $IDX/_plugins/_content_manager/rules/e2845de0-3719-400f-92c9-21881f3ef7b5
# promote draft -> test -> custom, then drop the integration and promote once more
curl -sk -u admin:admin -X DELETE $IDX/_plugins/_content_manager/integrations/235a7b5f-227b-4d7b-a460-6643db49ee56
```

```
== Summary ==
   28 passed / 0 failed
   ALL CHECKS PASSED
```

Cleanup leaves three `.opensearch-sap-log-types-config` documents behind
(`val1465-int-…` in `draft`/`test`/`custom`): the final promote races the
integration `DELETE`, the same visibility lag the script handles for rules with
`wait_rule_states` but not for the integration. Cosmetic — titles are
timestamp-suffixed, so repeat runs do not collide.

plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/DetectorLookupService.java
```diff
@@ -0,0 +1,241 @@
...
+                                if (parsed != null
+                                        && parsed.enabled()
+                                        && parsed.ruleIds().stream().anyMatch(ruleIds::contains)) {
+                                    detectors.add(parsed);
+                                }
+                            }
+                            listener.onResponse(detectors);
+                        },
+                        e -> {
+                            log.warn(Constants.W_LOG_DETECTOR_LOOKUP_FAILED, e.getMessage());
```
Jorgesnchz, line 142: We could do as is done in line 239, so it propagates failure and hard-fails the whole promotion, right now it swallows exceptions and returns an empty list

If not we can add a comment explaining it, if it is needed, but I think we could narrow the swallow to an absent index errors and let other errors propagate

Jorgesnchz (changes requested):
Marking it as requesting changes so https://github.com/wazuh/wazuh-indexer-plugins/pull/1465/changes#r3843340453 can be studied, but changes look good and validation too, just so we respond to the comment

If decided not to do anything I'll just approve

plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/DetectorLookupService.java
ferooo23: Fixed: 27bb3758f656fbb0ad10d4816da99025e018905c

Propagating like line 239, with the warn kept so the log says which lookup failed.

Jorgesnchz (approved):

PR #1165, opened 2026-05-13
Add index swap mechanism

Jorgesnchz:
### Description
This PR adds a new Index swap mechanism that allow us to maintain at all times the information in the indices clean for the user to see it correctly at every moment 

### Issues Resolved
IDR4799

PR #1474, opened 2026-08-24
Give the cluster time to go green in the integration tests

ferooo23:
## Description

`SpaceInitializationIT.testOnSyncCompleteDoesNotDuplicateSpaces` fails with `timed out waiting for
green state` on its first statement. The setup plugin creates around forty indices at node startup, one at a time, and each one takes the cluster through YELLOW, which outlasts the 30 second default of `ensureGreen()`. Nothing is unassigned, it is only slow (which is why it fails on some machines and not others).

Closes #1471

## Proposed Changes

`ensureGreen(TimeValue.timeValueMinutes(2))` on the two calls that open a test method and therefore
wait for the whole cluster to start:

- `SpaceInitializationIT.testOnSyncCompleteDoesNotDuplicateSpaces`
- `StandardSpaceHashRecoveryIT.testMissingStandardSpaceHashIsRecalculated`

The second one is included because the `integTest` cluster is shared by the whole task, so only the
first class to run pays the startup wait. 

The other two `ensureGreen()` calls in these classes are left alone: they wait for indices the test has just created, with the cluster already settled.

### Results and Evidence

Two minutes covers what I measured locally, with margin. It can be raised if CI needs it.

Before: `SpaceInitializationIT` fails at 30s with `timed out waiting for green state`.

After:

```
com.wazuh.contentmanager.SpaceInitializationIT       tests=1  failures=0  errors=0  time=87.555s
com.wazuh.contentmanager.StandardSpaceHashRecoveryIT tests=1  failures=0  errors=0  time=5.901s
```

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

None, this fixes existing tests.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Relevant evidence provided
- [ ] Tests cover the new functionality
- [ ] Configuration changes documented
- [ ] Developer documentation reflects the changes
- [ ] Meets requirements and/or definition of done
- [ ] No unresolved dependencies with other issues
- [ ] PR is linked to the relevant issue(s)
- [ ] Correct labels applied (e.g., `no-changelog`)

PR #1477, opened 2026-08-24
1476 wazuh threatintel enrichments a is never created so the cti writer auto creates a dynamically mapped index under the alias name and the home overviews threat catalog fails

### LINKED ISSUE #1476: `wazuh-threatintel-enrichments-a` is never created, so the CTI writer auto-creates a dynamically mapped index under the alias name and the home overview's Threat catalog fails
|---|---|---|---|---|
| 5.0.0-latest | wazuh-indexer-plugins (setup: threat-intel index provisioning), CTI content writer | Manager | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), single-node stack |

## Description

The indexer setup provisions six of the seven `wazuh-threatintel-*` indices as a concrete index named `<name>-a` with an alias named `<name>` pointing at it. `enrichments` is not in that set, so nothing creates it. The CTI content writer's first write therefore lands on a name that does not exist, OpenSearch auto-creates a concrete index literally named `wazuh-threatintel-enrichments`, and every field is dynamically mapped.

Every string field ends up as `text` with a `.keyword` subfield instead of the `keyword` the product declares. `text` has no doc_values, so aggregating or sorting on any of them is refused with a 400. On `/app/wz-home`, the page every login lands on, this takes out the Threat catalog widget group, which renders "Error fetching data".

The auto-created index also occupies the alias's name, so the alias cannot later be pointed at a properly created `-a` index without removing the squatter first.

## Screenshot

```
Home overview: could not load Threat intelligence

Threat intelligence: Error fetching data: search_phase_execution_exception:
[illegal_argument_exception] Reason: Text fields are not optimised for operations that require
per-document field data like aggregations and sorting, so these operations are disabled by
default. Please use a keyword field instead. Alternatively, set fielddata=true on [document.type]
in order to load field data by uninverting the inverted index. Note that this can use
significant memory.
```

Threat catalog widget group on the home overview, showing Error fetching data

## Current behavior

Six indices are provisioned correctly, the seventh is not:

```
wazuh-threatintel-decoders-a       -> alias wazuh-threatintel-decoders        OK
wazuh-threatintel-filters-a        -> alias wazuh-threatintel-filters         OK
wazuh-threatintel-integrations-a   -> alias wazuh-threatintel-integrations    OK
wazuh-threatintel-kvdbs-a          -> alias wazuh-threatintel-kvdbs           OK
wazuh-threatintel-policies-a       -> alias wazuh-threatintel-policies        OK
wazuh-threatintel-rules-a          -> alias wazuh-threatintel-rules           OK
wazuh-threatintel-enrichments      -> no alias: this is a concrete index      BROKEN
```

```
GET wazuh-threatintel-enrichments-a
  404 index_not_found_exception: no such index [wazuh-threatintel-enrichments-a]
```

The resulting mapping is the dynamic-mapping signature for a JSON string, not the mapping the product declares:

```json
"document.type": {"type": "text", "fields": {"keyword": {"type": "keyword", "ignore_above": 256}}}
```

`document.type` is the field the widget hits first, but the dynamic mapping applied to every string field, so 11 of the index's 42 fields refuse both aggregation and sorting:

```
document.feed.name   document.provider        document.software.name   document.tags
document.id          document.reference       document.software.type   document.type
document.name        document.software.alias  hash.sha256
```

Sorting is refused with the same message as aggregating, so a sortable column on any of them fails identically:

```
GET wazuh-threatintel-enrichments/_search  {"sort":[{"document.type":"asc"}]}
  400 search_phase_execution_exception / illegal_argument_exception
  "Text fields are not optimised for operations that require per-document field data
   like aggregations and sorting..." on [document.type]
```

`/app/wz-home` issues five search requests. Four succeed, the one against this index fails:

```
[ ok ] 200 wazuh-states-vulnerabilities*     vulnerabilities_by_package:package.name
[ ok ] 200 wazuh-states-fim-files*           fim_top_files:file.path
[ ok ] 200 wazuh-states-sca*                 sca_benchmarks:policy.name
[ ok ] 200 wazuh-findings-v5*                tactics, top_rules, techniques_count, top_techniques
[FAIL] 400 wazuh-threatintel-enrichments*    ioc_feed_by_type:document.type, threat_types:document.software.type
```

The rest of the page stays usable. Only this section is affected, but it fails as a unit, because both aggregations travel in one request issued by one hook:

```js
// wazuh.chunk.21.js
function useThreatIntelEnrichments(enabled) {          // label: "Threat intelligence"
  fetchData({aggs: {...buildThreatIntelFeedByTypeAgg(),      // terms on document.type
                    ...buildThreatIntelByThreatTypeAgg()},   // terms on document.software.type
             pagination: NO_HITS})
  return {total, feedByType, byThreatType}
}
```

`WidgetGroup title:"Threat catalog"` (`data-test-subj="home-overview-threat-catalog"`) renders `ThreatCatalogTiles{iocs}` and `ThreatTypeComposition{byThreatType}` from that return value. So "Threat catalog" is the widget group and "Threat intelligence" is the hook label that appears in the error banner, and the single 400 blanks three values at once: the IOC count tile, the feed-by-type breakdown and the threat-type composition.

The AI assistant reads the same index (`plugins/wazuhAiAssistant/server/tools/catalog/lookup-indicator.js`) but uses `sort: ['_doc']` and no aggregation, so its indicator lookup is unaffected.

## Expected behavior

`enrichments` is provisioned like its six siblings: a concrete `wazuh-threatintel-enrichments-a` with an explicit mapping and an alias `wazuh-threatintel-enrichments` pointing at it, with `document.type` mapped as an aggregatable `keyword`, and the Threat catalog widget group loading.

This is what a working deployment returns. The demo environment at `demo-5-0-0.xdrsiem.wazuh.info`, on the same dashboard build (`buildNumber 50005`), does not show the error and answers the same question with the correct shape:

```json
{"indices": ["wazuh-threatintel-enrichments-a"],
 "fields": {"document.type": {"keyword": {"type": "keyword", "searchable": true, "aggregatable": true}}}}
```

Note the resolved index name. There the `-a` index exists, the alias resolves to it, and the field is a `keyword`. The dashboard build and the query are identical on both, so the only difference is whether the index was created properly.

## Steps to reproduce

1. Deploy a single-node 5.0.0-latest stack from the public ECR `wazuh-cicd` images, with content initialization at its defaults (enabled).
2. Log in. The landing page `/app/wz-home` renders "Error fetching data" in the Threat catalog widget group, with a "could not load Threat intelligence" banner.
3. Confirm the cause against the indexer:

```bash
# the -a index is missing while its six siblings exist
curl -sk -u admin:admin 'https://localhost:9200/_cat/indices/wazuh-threatintel-*?h=index,creation.date.string,docs.count&s=creation.date'
curl -sk -u admin:admin 'https://localhost:9200/_alias/wazuh-threatintel-*'
curl -sk -u admin:admin 'https://localhost:9200/wazuh-threatintel-enrichments-a'      # 404

# the mapping that follows from that
curl -sk -u admin:admin 'https://localhost:9200/wazuh-threatintel-enrichments/_field_caps?fields=document.type'

# the user-visible failure, as an aggregation and as a sort
curl -sk -u admin:admin 'https://localhost:9200/wazuh-threatintel-enrichments/_search' \
  -H 'Content-Type: application/json' -d '{"size":0,"aggs":{"t":{"terms":{"field":"document.type","size":5}}}}'
curl -sk -u admin:admin 'https://localhost:9200/wazuh-threatintel-enrichments/_search?size=1' \
  -H 'Content-Type: application/json' -d '{"sort":[{"document.type":"asc"}]}'
```

### Checking any deployment from a browser

No indexer or SSH access needed. Open **Indexer management > Dev Tools** (`/app/dev_tools`; the one under Server management, `/app/dev-tools`, is the Wazuh API console and a different application) and run:

```
GET wazuh-threatintel-enrichments/_field_caps?fields=document.type
```

Read the `indices` array first, since it names what the alias resolved to:

- `["wazuh-threatintel-enrichments-a"]` with `"aggregatable": true`: healthy.
- `["wazuh-threatintel-enrichments"]` with `"aggregatable": false`: the `-a` index was never created and the widget group is broken.

## Root cause

The creation timestamps separate the two writers. The setup provisions six indices in a 1.1 second burst, each with its alias. The seventh appears 33 seconds later, un-suffixed, alias-less, and carrying the entire CTI corpus:

```
index                             created                    docs
wazuh-threatintel-decoders-a      2026-08-21T09:07:06.396Z     505
wazuh-threatintel-rules-a         2026-08-21T09:07:06.588Z     291
wazuh-threatintel-integrations-a  2026-08-21T09:07:06.883Z     125
wazuh-threatintel-filters-a       2026-08-21T09:07:07.151Z       0
wazuh-threatintel-kvdbs-a         2026-08-21T09:07:07.298Z      47
wazuh-threatintel-policies-a      2026-08-21T09:07:07.513Z       4
wazuh-threatintel-enrichments     2026-08-21T09:07:41.020Z  257070   <- 33s later, no -a, no alias
```

So `enrichments` is missing from whatever list drives the setup, and the CTI writer creates it implicitly instead.

### The intended mapping ships in the product and is never installed

This is what makes it a defect rather than a deployment choice. `wazuh-indexer-plugins` carries a setup template for exactly this index, declaring the field as a `keyword`:

`plugins/setup/src/main/resources/templates/content/ioc.json`

```json
{"index_patterns": ["wazuh-threatintel-enrichments*"],
 "priority": ...,
 "template": {"mappings": {"properties": {"document": {"properties": {
   "type": {"type": "keyword", "ignore_above": 1024}, ...}}}}}}
```

- Present and identical on `5.0.0`, on `main` and on the `v5.0.0-beta4` tag, so this is not an unreleased change the image predates.
- Extracting `wazuh-indexer-setup-5.0.0.latest.jar` from the running indexer container shows `templates/content/ioc.json` inside, so this is not version skew between container and source.

The dashboard declares the same intent independently, which is why it aggregates the raw field name rather than falling back to `document.type.keyword`. `plugins/wazuh/common/known-fields/threatintel-enrichments.json`, which seeds the index pattern's field list, states:

```
document.type            type=string  aggregatable=true  searchable=true
document.software.type   type=string  aggregatable=true  searchable=true
document.name            type=string  aggregatable=true  searchable=true
document.provider        type=string  aggregatable=true  searchable=true
```

The dashboard is therefore not doing anything wrong: it is told the field is aggregatable, and the product's own template says the same. Only the index disagrees. Note also that `wazuh-threatintel-enrichments*` is the only index pattern registered for the whole threat-intel family, out of 46 in the deployment, so it is the one every dashboard consumer of CTI content goes through. The six healthy siblings have no index pattern at all.

The template is simply never installed. Shipped versus installed, by group:

```
group      shipped                                        installed
content     2  (filters, ioc)                              0 / 2     <- neither
states     20  (fim-files, inventory-*, sca, ...)          20 / 20
streams     8  (events, findings, metrics-*, raw, ...)       8 / 8
root        4  (cve, ism-config, settings, setup-status)     2 / 4

templates in the cluster: 45      covering any wazuh-threatintel-* index: NONE
```

One code path is broken and the others are not: the setup installs every state and stream template it carries, and neither of its two content templates.

Note that a template is not the only route to a correct mapping. The six healthy siblings have no template either, they were created explicitly with their mappings. The accurate statement is that the setup provisioned nothing for this content type, neither its template nor its `-a` index and alias, while it provisioned both aspects for the other six.

## Additional impact: the same failure returned as success

Aggregating the same field across the family does not error, it returns a wrong answer. One shard of seven fails and the other six succeed, so the partial result is reported as a success:

```
GET wazuh-threatintel-*/_search  {"aggs":{"t":{"terms":{"field":"document.type"}}}}

_shards: {"total": 7, "successful": 6, "failed": 1,
          "failures": [{"index": "wazuh-threatintel-enrichments", ...}]}
buckets: []
```

A caller that does not inspect `_shards.failed` gets empty buckets and no error at all.

### LINKED ISSUE #1481: VD consumer reports ready while OSCPE-GLOBAL is missing; manager polls forever and package inventory stays empty
|---|---|---|---|---|
| 5.0.0-latest | wazuh-manager (content-updater `IndexerDownloader`, vulnerability-scanner), wazuh-agent (syscollector) | Manager and agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), single-node stack |

## Summary

On a fresh single-node stack, one of the three global-map documents the vulnerability scanner requires — `OSCPE-GLOBAL` — never lands in `.wazuh-threatintel-vulnerabilities`. The CTI consumer nevertheless reports `status: ready` with `local_offset == remote_offset`, so nothing will ever re-download it. The manager polls for the document every 30 seconds, at INFO, indefinitely.

Because agents will not synchronize their package inventory until the server reports a VD feed offset, the blast radius is wider than vulnerability detection: `wazuh-states-inventory-packages` also stays empty and the IT Hygiene package views are blank on a stack where every other inventory table is populated.

The chain, end to end, is fully in the logs:

> disk high watermark → `cluster.blocks.create_index` → declared index creation refused and never retried → bulk write auto-creates the index with inferred mappings → inferred mappings reject `OSCPE-GLOBAL` → consumer reports `ready` → manager waits forever → no packages, no vulnerabilities.

### 1. The consumer considers itself fully synchronized

```json
{
  "name": "public-vulnerabilities-5",
  "context": "t1-vulnerabilities-5",
  "type": "cti:catalog:consumer:vulnerabilities",
  "status": "ready",
  "local_offset": 849527,
  "remote_offset": 849527
}
```

### 2. Two of the three required global maps are present

The content is largely there. Document counts by type:

```
$ POST .wazuh-threatintel-vulnerabilities/_search  (terms on type.keyword)
  372113  CVE
     181  TID
       1  CNA-MAPPING-GLOBAL
       1  FEED-GLOBAL
```

The three names the content manager looks for are in the library itself:

```
$ grep -aoE '[A-Z][A-Z0-9]*(-[A-Z0-9]+)*-GLOBAL' /var/wazuh-manager/lib/libcontent_manager.so | sort -u
CNA-MAPPING-GLOBAL
FEED-GLOBAL
OSCPE-GLOBAL
```

`OSCPE-GLOBAL` is the missing third:

```
$ POST .wazuh-threatintel-vulnerabilities/_count {"query":{"term":{"type.keyword":"OSCPE-GLOBAL"}}}
{"count":0}
```

### 3. The manager loops on it indefinitely, at INFO

<details>
<summary>modulesd content-updater log</summary>

```
2026/08/25 08:31:05 wazuh-manager-modulesd:content-updater: INFO: IndexerDownloader: Global-map documents not yet indexed in '.wazuh-threatintel-vulnerabilities' (2/3 found). Waiting 30...
2026/08/25 08:31:35 ... (2/3 found). Waiting 30...
2026/08/25 08:32:05 ... (2/3 found). Waiting 30...
```

</details>

The scanner had already given up at startup:

```
07:15:49 wazuh-manager-modulesd:vulnerability-scanner: The local feed database is incomplete at startup: required global maps are missing. Clearing update...
07:15:49 wazuh-manager-modulesd:vulnerability-scanner: CVE feed not yet available, per-agent scans deferred until initial load completes.
```

### 4. Agents defer package sync on a five-minute cycle, indefinitely

```
2026/08/25 08:31:07 wazuh-modulesd:syscollector: INFO: Syscollector VD synchronization deferred: No VD feed offset available yet. Waiting for the server to report one via /control. Cannot proceed with VD synchronization.
```

Visible result on a stack whose other inventory tables are healthy:

```
wazuh-states-inventory-packages    0      <- gated behind the VD sync
wazuh-states-vulnerabilities       0
wazuh-states-inventory-users      28
wazuh-states-inventory-processes  21
wazuh-states-fim-files          1598
wazuh-states-sca                 366
```

### The index being polled is mis-provisioned

`.wazuh-threatintel-vulnerabilities` was created without the `-a` suffix and without an alias, unlike every content index created earlier in the same startup:

```
07:13:08  wazuh-threatintel-integrations-a     alias -> wazuh-threatintel-integrations
07:13:09  wazuh-threatintel-rules-a            alias -> wazuh-threatintel-rules
07:13:09  wazuh-threatintel-decoders-a         alias -> wazuh-threatintel-decoders
07:13:09  wazuh-threatintel-policies-a         alias -> wazuh-threatintel-policies
07:13:10  wazuh-threatintel-kvdbs-a            alias -> wazuh-threatintel-kvdbs
07:13:10  wazuh-threatintel-filters-a          alias -> wazuh-threatintel-filters
--- cluster.blocks.create_index goes up somewhere in here, see wazuh/wazuh#36 ---
07:13:43  wazuh-threatintel-enrichments        no -a, no alias, dynamic mappings
07:15:01  .wazuh-threatintel-vulnerabilities   no -a, no alias, dynamic mappings
```

Same squatting pattern as wazuh/wazuh#10, with the same signature: `type` was inferred as `text` rather than `keyword`, so aggregating on it throws `illegal_argument_exception`, and an exact term query for `"CVE"` matches nothing while `"cve"` matches all 372,113 documents.

### The document was sent, and the indexer refused it

The missing document is a consequence of the mis-provisioning, not a gap in the feed. `OSCPE-GLOBAL` was delivered and rejected at index time:

```
[2026-08-25T07:15:55,370][WARN][ContentIndex] Bulk indexing finished with failures:
[66]: index [.wazuh-threatintel-vulnerabilities], id [OSCPE-GLOBAL], message
      [java.lang.IllegalArgumentException: can't merge a non object mapping
       [document.Microsoft Windows 8] with an object mapping]
```

It is not alone — the same bulk runs lost other documents to the same class of conflict:

```
[0]:   id [CVE-1999-1121]  mapper [document.containers.cna.metrics.cvssV2_0.baseScore]
                           cannot be changed from type [long] to [float]
[112]: id [TCPE]           object field starting or ending with a [.] makes object resolution ambiguous
```

Every one of these is a mapping conflict, i.e. what happens when an index infers its mappings from the first documents it sees instead of being created with a declared mapping. This index has inferred mappings because its proper creation was refused:

```
07:14:01,332  AbstractConsumerService  Failed to create index [.wazuh-threatintel-vulnerabilities]:
                                       IndexCreateBlockException[blocked by: [FORBIDDEN/10/cluster create-index blocked (api)]]
07:15:01,619  MetadataCreateIndexService  [.wazuh-threatintel-vulnerabilities] creating index, cause [auto(bulk api)]
```

The declared creation is blocked, the code carries on, and the next bulk write auto-creates the index with whatever mappings the data implies. Full sequence: [`evidence/02-logs/EXTRACT-race-create-index-block.log`](evidence/02-logs/EXTRACT-race-create-index-block.log). The block itself is wazuh/wazuh#36.

> [!NOTE]
> The `cluster.blocks.create_index` block is set by OpenSearch's `DiskThresholdMonitor` because the host disk crossed the high watermark — it is not set by Wazuh. See the correction at the top of wazuh/wazuh#5.
>
> The chain and the product defects in it are unchanged: the declared index creation is refused and never retried, the bulk write squats the name with inferred mappings, those mappings reject `OSCPE-GLOBAL`, and the manager then waits for it forever while reporting `status: ready`. A disk briefly filling up should not cost a deployment its vulnerability detection and its package inventory permanently.

The manager's side of this is a defect on its own: whatever the reason a mandatory document is absent, reporting `status: ready` and then polling forever for it is not a reasonable response.

## Expected behavior

- A consumer that has reached `remote_offset` and cannot satisfy a mandatory document should fail loudly rather than report `ready` and poll forever. `status: ready` currently means "I downloaded everything I was offered", not "the feed is usable", and nothing distinguishes the two.
- Waiting for a document that no further download can produce is a deadlock, not a wait. Once the consumer is complete, the manager should re-request the missing document or restart the sync, and report the condition at WARN or ERROR rather than INFO.
- The package inventory should not be gated on the vulnerability feed. Losing package visibility because a CVE feed is incomplete is a much larger failure than the one that caused it.

## Steps to reproduce

Observed on a fresh single-node stack; reproducible for as long as the stack is up.

1. Deploy the stack and enroll an agent.
2. Confirm the consumer reports itself done — `status: ready`, `local_offset == remote_offset`:
   ```bash
   curl '.wazuh-cti-consumers/_doc/cti:catalog:consumer:vulnerabilities'
   ```
3. Confirm the mandatory document is absent — returns `0`:
   ```bash
   curl '.wazuh-threatintel-vulnerabilities/_count' \
     -d '{"query":{"term":{"type.keyword":"OSCPE-GLOBAL"}}}'
   ```
4. Observe the manager log repeating `(2/3 found). Waiting 30...` indefinitely.
5. Observe `wazuh-states-inventory-packages` and `wazuh-states-vulnerabilities` staying at 0 while the other inventory tables fill.

felipegonzalezmv:
## Description

Six of the seven `wazuh-threatintel-*` indices are provisioned as `<name>-a` behind a `<name>` alias.
`enrichments` was not, so nothing created it. The first CTI write therefore hit a name that did not
exist, OpenSearch auto-created a concrete `wazuh-threatintel-enrichments` with dynamic mappings, and
every declared `keyword` became `text` — which has no doc_values.

That breaks the **Threat catalog** group on `/app/wz-home`: its two aggregations (`document.type`,
`document.software.type`) travel in one request, so one 400 blanks all three tiles. The squatting
index also occupies the alias's name, so the alias cannot be repointed until it is removed.

Per [this comment](https://github.com/wazuh/wazuh-indexer-plugins/issues/1476#issuecomment-5394842395),
creation of the whole family moves to the setup plugin.

Closes #1476

## Proposed Changes

**Setup plugin — now owns the threat-intel indices**
- New `ContentIndex` kind (third `WazuhIndex` subclass): creates `<name>-a` + write alias `<name>`,
  and deletes any concrete index squatting the alias name. Template patterns become `<name>*` so they
  also cover the `-b` shadow slot the content manager swaps into.
- All 8 indices registered (7 `wazuh-threatintel-*` + hidden `.wazuh-threatintel-vulnerabilities`),
  with 6 new templates under `templates/content/`.
- Added the missing `offset: unsigned_long` to `content/ioc.json`. Every IOC carries it and the
  template is `dynamic: strict_allow_templates`, so shipping it as-is would have rejected every CTI
  write. Also added to the WCS `content/ioc` module so regeneration keeps it.

**Content manager — stops racing and stops auto-creating**
- `ensureResourceIndicesExist()` now waits for the setup readiness marker. Registering the indices in
  setup was not enough: this method waited for nothing and won, creating 6 of the 8 *before* their
  templates existed. Reuses the existing wait, extracted to `SetupReadiness`.
- A failed index creation now aborts the sync pass instead of letting the next bulk auto-create a
  dynamically mapped index — the root cause of this bug.
- An IOC/CVE index found empty with a non-zero offset resets the offset, so a removed squatter's
  content is re-downloaded.

**Guard against drift** — the mappings deliberately live in both plugins (the content manager needs
its copies for shadow indices and standalone runs). New `verifyContentTemplates` Gradle task, wired
into `check`, fails the build if they diverge. It would already have caught the missing `offset`.

### Results and Evidence

Verified on a real node built from the ARCHIVE distribution with this branch's plugin zips.

**Before / after**, reproduced on the same node:

| | before | after |
|---|---|---|
| index | `wazuh-threatintel-enrichments` (concrete) | `wazuh-threatintel-enrichments-a` |
| alias | `404 alias [...] missing` | `-> ...-a`, `is_write_index: true` |
| `document.type` | `{"text": {"aggregatable": false}}` | `{"keyword": {"aggregatable": true}}` |
| Threat catalog aggs | `400` *Text fields are not optimised…* | `200`, 0 failed shards |

`_field_caps` now returns byte-for-byte what the issue quotes from the healthy demo environment:

```json
{"indices": ["wazuh-threatintel-enrichments-a"],
 "fields": {"document.type": {"keyword": {"type": "keyword", "searchable": true, "aggregatable": true}}}}
```

**Setup is the creator.** With the full stack installed, every index names its own template — an
empty `templates []` is the signature of an index created before its template existed:

```
wazuh-threatintel-enrichments-a        template=[wazuh-threatintel-enrichments-template]
wazuh-threatintel-filters-a            template=[wazuh-threatintel-filters-template]
...all 8...
```

The content manager provisioned none of them. Without the readiness wait, 6 read `templates []`.

**Upgrade is self-healing** — restarting over the broken state:

```
[WARN][c.w.s.i.ContentIndex] Found a concrete index named [wazuh-threatintel-enrichments], which is the
  name reserved for the public alias. ... Deleting it so the alias can be created.
[INFO][c.w.s.i.ContentIndex] Index created successfully: wazuh-threatintel-enrichments-a true (alias ...)
```

**Build** — `./gradlew check -PnumNodes=1` from each plugin dir, as CI does. All green:
setup 55 unit / 28 integ, content-manager 584 unit / 149 integ. 0 failures, 0 skipped.

### Artifacts Affected

- `wazuh-indexer-setup`: new `ContentIndex`, 8 registered indices, 7 bundled templates (6 new,
  1 amended), new `verifyContentTemplates` task.
- `wazuh-indexer-content-manager`: new `SetupReadiness`, readiness wait, sync aborts instead of
  auto-creating, offset reset for single-index consumers.
- WCS `content/ioc`: `offset` added to `fields/custom/base.yml` and `fields/subset.yml`.

On a fresh cluster this adds `wazuh-threatintel-enrichments-a` and
`.wazuh-threatintel-vulnerabilities-a`, plus 8 `<alias>-template` templates.

### Configuration Changes

None — no settings or defaults changed.

**Nothing was renamed.** The `-a`/`-b` scheme predates this PR (`ContentIndex.SUFFIX_A`, #1165) and 6
of 7 indices already had it. Alias names are unchanged and the 8 index-name constants are identical to
`main`; only the concrete index behind the alias changes, and only for `enrichments`, which had none.

**Upgrades** need no manual step: setup deletes the squatter and creates `-a` + alias, the content
manager resets the offset and re-downloads from CTI. The deleted content is CTI-sourced.

### Documentation Updates

`docs/dev/plugins/setup.md` — `ContentIndex` added to the class diagram and the list of index kinds,
plus a new "Threat intel content indices" section (layout and why the alias exists, which plugin
creates them and how to confirm it, squatter recovery, why mappings live in both plugins). Also notes
that `templates/cve.json` is a WCS artifact for `wazuh-cve*` and intentionally unregistered.

`CHANGELOG.md` — entry under `## [v5.0.0]` → `### Fixed`.

### Tests Introduced

- **`ThreatIntelIndicesIT`** (6) — all 8 `-a` indices, aliases and templates exist; plus the issue's
  own diagnostic as assertions: `document.type` aggregatable on `...-a`, the Threat catalog agg pair
  returns 200 with no failed shards, and an IOC with `offset` is accepted.
- **`ContentIndexTests`** (6) — creation with alias, squatter deleted first, alias added when only the
  concrete index exists, no-op when both exist, `ResourceAlreadyExistsException` swallowed.
- **`SetupReadinessTests`** (5) — ready / failed / running / unreadable marker, and the full backoff
  schedule exhausted before giving up.
- **`verifyContentTemplates`** — verified in both directions (passes clean, fails on injected drift)
  and in the standalone `cd plugins/setup && ./gradlew check` layout CI uses.
- **Two existing ITs fixed, not worked around.** `SpaceInitializationIT` and
  `StandardSpaceHashRecoveryIT` created concrete indices named after the aliases setup now provisions;
  both reach the Gradle cluster as an `ExternalTestCluster`, so the setup plugin is running there.
  Confirmed this PR caused it, not flakiness: `main` 3/3 pass, branch 2/2 fail, and with
  content-manager reverted to `main` it still failed 3/3 — isolating the cause to the setup side.

## Review Checklist

- [x] Code changes reviewed
- [x] Configuration changes documented (none introduced)
- [ ] Verified on a full 5.0.0 stack (Docker/ECR) per the issue's reproduction steps

### Two points for reviewers

1. **The readiness wait reuses the catalog-download budget** (300s default). With the setup plugin
   absent, custom-ruleset endpoints would 503 for that window before the fallback runs. Both plugins
   always ship together, so this only affects setup-less dev clusters. A shorter budget for this gate
   would be a separate setting.
2. **Squatter deletion.** Lossless for `enrichments` and CVE (CTI-sourced). For the 6 ruleset indices
   a squatter is effectively impossible — their aliases exist from first boot, before any REST
   endpoint is reachable — but if one existed it could hold user-authored rules, which are not
   re-downloadable. No guard added; happy to log the discarded doc count or reindex instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,7 @@
 - [BUG] Undefined state if a node is restarted while the standard space hash is being calculated [(#1773)](https://github.com/wazuh/wazuh-indexer/issues/1773)
 - [BUG] Fix issues in workflows [(#1449)](https://github.com/wazuh/wazuh-indexer/issues/1449)
 - Failed to save/update ManagedIndexMetaData [(#1828)](https://github.com/wazuh/wazuh-indexer/issues/1828)
+- [BUG] `wazuh-threatintel-enrichments-a` is never created, so the CTI writer auto-creates a dynamically mapped index under the alias name and the home overview's Threat catalog fails [(#1476)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1476)
 
 ## Prior versions
 - []()
--- a/docs/dev/plugins/setup.md
+++ b/docs/dev/plugins/setup.md
@@ -21,6 +21,7 @@ classDiagram
     <<abstract>> WazuhIndex
     class StateIndex
     class StreamIndex
+    class ContentIndex
 
     %% Relations
     IndexInitializer <|-- Index : implements
@@ -28,6 +29,7 @@ classDiagram
     Index <|-- WazuhIndex
     WazuhIndex <|-- StateIndex
     WazuhIndex <|-- StreamIndex
+    WazuhIndex <|-- ContentIndex
 
     %% Schemas
     class IndexInitializer {
@@ -63,6 +65,15 @@ classDiagram
         +createIndex(String index)
     }
     class StateIndex {
+    }
+    class ContentIndex {
+        +String SUFFIX_A
+        +ContentIndex(String index, String template)
+        +getPhysicalName() String
+        +createIndex(String index)
+        +createTemplate(String template)
+        -deleteSquatterIndex(String alias) void
+        -ensureAlias(String alias, String physicalName) void
     }
 ```
 
@@ -144,10 +155,11 @@ Follow the existing structure and naming convention. Example:
 
 Edit the constructor of the `SetupPlugin` class located at: `/plugins/setup/src/main/java/com/wazuh/setup/SetupPlugin.java`
 
-Add the template and index entry to the `indices` map. There are two kinds of indices:
+Add the template and index entry to the `indices` map. There are three kinds of indices:
 
 - **Stream index**. Stream indices contain time-based events of any kind (alerts, statistics, logs...). These are created as Data Streams.
 - **Stateful index**. Stateful indices represent the most recent information of a subject (active vulnerabilities, installed packages, open ports, ...). These indices are different from Stream indices as they do not contain timestamps. The information is not based on time, as they always represent the most recent state.
+- **Content index**. Content indices hold the CTI catalog content the Content Manager downloads. They are not addressed by their own name: see [Threat intel content indices](#threat-intel-content-indices) below.
 
 ```java
 /**
@@ -162,6 +174,8 @@ public class SetupPlugin extends Plugin implements ClusterPlugin {
   this.indices.add(new StreamIndex("my-stream-index", "templates/streams/my-index-template-1"));
   // State indices
   this.indices.add(new StateIndex("my-state-index", "templates/states/my-index-template-2"));
+  // Content indices
+  this.indices.add(new ContentIndex("my-content-index", "templates/content/my-index-template-3"));
 
   //...
 }
@@ -511,6 +525,78 @@ The **stream-metrics-policy** manages all `wazuh-metrics-*` data streams (`wazuh
 2. **Delete State**
    - Actions: Deletes the index
    - Retry Policy: 3 attempts with exponential backoff (1-minute initial delay)
+
+---
+
+## Threat intel content indices
+
+### Overview
+
+The `wazuh-threatintel-*` indices hold the CTI catalog content the Content Manager downloads:
+IoCs, CVEs and the Engine ruleset. The setup plugin provisions all of them.
+
+| Alias | Concrete index | Template |
+| --- | --- | --- |
+| `wazuh-threatintel-enrichments` | `wazuh-threatintel-enrichments-a` | `templates/content/ioc.json` |
+| `wazuh-threatintel-filters` | `wazuh-threatintel-filters-a` | `templates/content/filters.json` |
+| `wazuh-threatintel-decoders` | `wazuh-threatintel-decoders-a` | `templates/content/decoders.json` |
+| `wazuh-threatintel-rules` | `wazuh-threatintel-rules-a` | `templates/content/rules.json` |
+| `wazuh-threatintel-kvdbs` | `wazuh-threatintel-kvdbs-a` | `templates/content/kvdbs.json` |
+| `wazuh-threatintel-integrations` | `wazuh-threatintel-integrations-a` | `templates/content/integrations.json` |
+| `wazuh-threatintel-policies` | `wazuh-threatintel-policies-a` | `templates/content/policies.json` |
+| `.wazuh-threatintel-vulnerabilities` | `.wazuh-threatintel-vulnerabilities-a` | `templates/content/vulnerabilities.json` |
+
+### Why the alias
+
+A content index is never addressed by its own name. `ContentIndex` creates the concrete index as
+`<name>-a` and points a write alias named `<name>` at it. The Content Manager needs that
+indirection: when the content source changes (a plan upgrade or downgrade), it rebuilds the whole
+catalog into a hidden `<name>-b` shadow index and swaps the alias in a single atomic action, so
+readers never see a partially populated index. The two slots alternate on each swap.
+
+`ContentIndex.createTemplate()` therefore rewrites the template's `index_patterns` to `<name>*`,
+whatever the shipped JSON declares. That covers the live slot, the shadow slot and — as a last line
+of defence — the alias name itself.
+
+### Recovering an index that squats the alias name
+
+If a write reaches the alias name before anything creates the alias, OpenSearch auto-creates a
+concrete index called `<name>` with dynamic mappings. Every keyword the schema declares becomes a
+`text` field, which has no doc_values and so cannot be aggregated or sorted on, and the index
+occupies the name the alias needs. `ContentIndex.createIndex()` deletes such an index before
+creating `<name>-a`. The Content Manager then finds the index empty with a non-zero local offset and
+resets the offset, so the content is downloaded again on the same synchronization pass.
+
+### Where the mappings live
+
+The mappings exist in two places on purpose. This plugin ships them as index templates under
+`plugins/setup/src/main/resources/templates/content/`, and the Content Manager keeps its own copies
+under `plugins/content-manager/src/main/resources/mappings/cti-*-mappings.json`, which it needs to
+create the shadow indices and to run without this plugin installed. The
+`verifyContentTemplates` Gradle task, wired into `check`, fails the build if the two sets diverge.
+
+`templates/content/ioc.json` and `templates/content/filters.json` are generated by the WCS
+generator; the other six are maintained by hand alongside the Content Manager copies.
+`templates/cve.json` is **not** the CVE content template: it is a WCS artifact for the unrelated
+`wazuh-cve*` pattern and is intentionally not registered.
+
+### Testing
+
+```bash
+./gradlew :wazuh-indexer-setup:integTest --tests '*ThreatIntelIndicesIT*'
+```
+
+Against a deployed cluster, the alias must resolve to the `-a` index and its fields must be
+aggregatable:
+
+```bash
+curl -sk -u admin:admin 'https://localhost:9200/_alias/wazuh-threatintel-*'
+curl -sk -u admin:admin \
+  'https://localhost:9200/wazuh-threatintel-enrichments/_field_caps?fields=document.type'
+```
+
+The `indices` array in the `_field_caps` response names what the alias resolved to. It must read
+`["wazuh-threatintel-enrichments-a"]` with `"aggregatable": true`.
 
 ---
 
--- a/docs/dev/plugins/setup.md
+++ b/docs/dev/plugins/setup.md
@@ -567,6 +567,19 @@ present. Without that wait it won the race by a wide margin — it created six o
 seconds before this plugin reached its threat-intel block, which meant they were created before
 their index templates were installed and so did not pick them up.
 
+Downloading CTI content requires two things to be true, and the Content Manager checks both before
+any write reaches these indices:
+
+1. `.wazuh-setup-status` reports `status: ready`, which this plugin writes only after every index it
+   owns exists.
+2. The target indices of that consumer exist.
+
+If either check fails the consumer skips its pass and logs at ERROR; it creates nothing. That is
+what keeps a mis-provisioned index from becoming a squatted one: a write through the alias name
+would otherwise auto-create a concrete index there with dynamic mappings. See issues #1476 and
+#1481, where the second of those took out vulnerability detection and package inventory because the
+global-map documents could not be indexed into an inferred mapping.
+
 To confirm on a running cluster, read the `templates` field of the creation entries in the log:
 
 ```bash
--- a/docs/dev/plugins/setup.md
+++ b/docs/dev/plugins/setup.md
@@ -558,6 +558,24 @@ readers never see a partially populated index. The two slots alternate on each s
 whatever the shipped JSON declares. That covers the live slot, the shadow slot and — as a last line
 of defence — the alias name itself.
 
+### Who creates them
+
+This plugin does, and only this plugin, on a cluster where both are installed. The Content Manager
+keeps an `ensureResourceIndicesExist()` fallback so a deployment without this plugin still works,
+but it waits for the `.wazuh-setup-status` readiness marker first and then finds the indices already
+present. Without that wait it won the race by a wide margin — it created six of the eight about 16
+seconds before this plugin reached its threat-intel block, which meant they were created before
+their index templates were installed and so did not pick them up.
+
+To confirm on a running cluster, read the `templates` field of the creation entries in the log:
+
+```bash
+grep 'creating index' /var/log/wazuh-indexer/wazuh-cluster.log | grep threatintel
+```
+
+Every `wazuh-threatintel-*-a` line must name its own `<name>-template`. An empty `templates []`
+means something created the index before the template existed.
+
 ### Recovering an index that squats the alias name
 
 If a write reaches the alias name before anything creates the alias, OpenSearch auto-creates a
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/ContentManagerPlugin.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/ContentManagerPlugin.java
@@ -592,7 +592,9 @@ public class ContentManagerPlugin extends Plugin
      * Setup had even installed their index templates (see issue #1476).
      */
     private void ensureResourceIndicesExist() {
-        if (!this.setupReadiness.awaitReady()) {
+        // Only a Setup plugin that is installed but did not become ready is worth warning about.
+        // Its absence is the expected path for this fallback and awaitReady() already reports it.
+        if (!this.setupReadiness.awaitReady() && this.setupReadiness.isSetupPluginInstalled()) {
             log.warn(Constants.W_LOG_SETUP_NOT_READY_PROVISIONING);
         }
         for (Map.Entry<String, String> entry : Constants.RESOURCE_INDEX_MAPPINGS.entrySet()) {
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/ContentManagerPlugin.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/ContentManagerPlugin.java
@@ -139,6 +139,7 @@ public class ContentManagerPlugin extends Plugin
     private final AtomicBoolean standardSpaceHashInFlight = new AtomicBoolean(false);
     private SpaceService spaceService;
     private SecurityAnalyticsService securityAnalyticsService;
+    private SetupReadiness setupReadiness;
     private Environment environment;
     private ClusterService clusterService;
     private LogtestService logtestService;
@@ -201,6 +202,7 @@ public class ContentManagerPlugin extends Plugin
 
         this.consumersIndex = new ConsumersIndex(client);
         this.credentialsIndex = new CredentialsIndex(client, threadPool);
+        this.setupReadiness = new SetupReadiness(client);
         this.plansService = new PlansServiceImpl();
         this.subscriptionService =
                 new SubscriptionServiceImpl(
@@ -505,22 +507,11 @@ public class ContentManagerPlugin extends Plugin
                                                 e);
                                     }
 
-                                    // The Setup plugin owns the threat-intel index topology: each one
-                                    // is created as <name>-a with the public alias <name> pointing at
-                                    // it. Wait for it to report readiness before touching them, so this
-                                    // fallback does not race it (see issue #1476) — without the wait it
-                                    // wins the race and the indices are created here, from this
-                                    // plugin's own mappings, before Setup has even installed their
-                                    // index templates.
-                                    if (!new SetupReadiness(this.client).awaitReady()) {
-                                        log.warn(Constants.W_LOG_SETUP_NOT_READY_PROVISIONING);
-                                    }
-
                                     // Create the threat-intel ruleset resource indices up front so
                                     // custom-ruleset REST endpoints work even when catalog
                                     // synchronization is disabled, or when the Setup plugin is not
-                                    // installed at all. Whenever Setup did provision them, the
-                                    // existence check below makes this a no-op.
+                                    // installed at all. Waits for the Setup plugin first; see the
+                                    // method's documentation.
                                     this.ensureResourceIndicesExist();
 
                                     // Seed the default space policies (draft, test, custom) so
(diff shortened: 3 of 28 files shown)


Jorgesnchz (changes requested):
Reviewed the changes and found some things:

- Since we squash and merge the name of the PR becomes visible as the commit name for the complete PR, example: `Give the cluster time to go green in the integration tests #1474`, this PR name is really long and would be too long for the GitHub commits history

- Same problem with the branch, we usually use: `bug/<issue-nº>-...` or `enhancement/<issue-nº>-...`, right now the branch name is a really long one and not following the format

- The email checker is failing: https://github.com/wazuh/wazuh-indexer-plugins/actions/runs/32884715694/job/97922343341, this is because we use our wazuh mail to sign the commits

- I believe that the main change of the issue and the best win for our side will be to migrate all the creation to WCS modules, since that will make maintenance way easier, IMO since we already have the `/content` folder we can create the additional modules

- Related to the modules, right now we have: The files in setup plugin (`plugins/setup/src/main/resources/templates/content/*`) and the ones in content manager (`plugins/content-manager/src/main/resources/mappings`), IMO we should make it that the WCS module loads both and updates them, so we can make sure that if we apply one change is applied in both and avoid errors (We maintain both, so then the CM can perform the swap without setup intervention)

- `ensureResourceIndicesExist` creates some of the resources, not all of them, so IOCs or CVEs will fail, since CM is highly connected to setup I believe that we can avoid this, if setup is not installed then it fails

The rest looks good, @AlexRuiz7 what do you think about it, maybe this is not this issue scope to create the WCS modules, and I'm asking for things out of the scope, you call the shots

PR #1486, opened 2026-08-26
Create the threat-intel content indices in the setup plugin

felipegonzalezmv:
## Description

Six of the seven `wazuh-threatintel-*` indices are provisioned as `<name>-a` behind a `<name>` alias.
`enrichments` was not, so nothing created it. The first CTI write therefore hit a name that did not
exist, OpenSearch auto-created a concrete `wazuh-threatintel-enrichments` with dynamic mappings, and
every declared `keyword` became `text` — which has no doc_values.

That breaks the **Threat catalog** group on `/app/wz-home`: its two aggregations (`document.type`,
`document.software.type`) travel in one request, so one 400 blanks all three tiles. The squatting
index also occupies the alias's name, so the alias cannot be repointed until it is removed.

Per [this comment](https://github.com/wazuh/wazuh-indexer-plugins/issues/1476#issuecomment-5394842395),
creation of the whole family moves to the setup plugin.

Closes #1476

## Proposed Changes

**Setup plugin — now owns the threat-intel indices**
- New `ContentIndex` kind (third `WazuhIndex` subclass): creates `<name>-a` + write alias `<name>`,
  and deletes any concrete index squatting the alias name. Template patterns become `<name>*` so they
  also cover the `-b` shadow slot the content manager swaps into.
- All 8 indices registered (7 `wazuh-threatintel-*` + hidden `.wazuh-threatintel-vulnerabilities`),
  with 6 new templates under `templates/content/`.
- Added the missing `offset: unsigned_long` to `content/ioc.json`. Every IOC carries it and the
  template is `dynamic: strict_allow_templates`, so shipping it as-is would have rejected every CTI
  write. Also added to the WCS `content/ioc` module so regeneration keeps it.

**Content manager — stops racing and stops auto-creating**
- `ensureResourceIndicesExist()` now waits for the setup readiness marker. Registering the indices in
  setup was not enough: this method waited for nothing and won, creating 6 of the 8 *before* their
  templates existed. Reuses the existing wait, extracted to `SetupReadiness`.
- A failed index creation now aborts the sync pass instead of letting the next bulk auto-create a
  dynamically mapped index — the root cause of this bug.
- An IOC/CVE index found empty with a non-zero offset resets the offset, so a removed squatter's
  content is re-downloaded.

**Guard against drift** — the mappings deliberately live in both plugins (the content manager needs
its copies for shadow indices and standalone runs). New `verifyContentTemplates` Gradle task, wired
[...]

PR #1391, opened 2026-07-16
Keep last working snapshot

f-galland:
## Description

The content-manager plugin's catalog synchronization pipeline suffered from excessive memory pressure, lack of resilience on partial failures, and a missing rollback path when a remote snapshot download failed after the previous working snapshot had already been discarded. This PR addresses all three areas.

Closes #1350 #1383

## Proposed Changes

- **Keep last working snapshot ("stable snapshot")**: After a successful snapshot download and index, the ZIP is renamed to `*.stable.zip` instead of being deleted. If a future remote download fails, the service falls back to re-indexing from this stable copy. A new `TransportPromoteSnapshotAction` (cluster-wide `TransportNodesAction`) broadcasts the promote operation to every node so all nodes retain their own local stable copy.
- **Per-offset checkpoint and resumable updates**: `UpdateServiceImpl` now persists a lightweight checkpoint to the consumers index after every batch (using `WriteRequest.RefreshPolicy.NONE`). On failure the consumer offset is left at the last checkpoint instead of being reset to 0, allowing the next sync cycle to resume from where it left off rather than replaying the entire update range.
- **Batched updates with MultiGet + BulkRequest**: UPDATE operations are accumulated into sub-batches of 50, fetched via a single `MultiGetRequest`, patched in-memory, and written back in a single `BulkRequest`. An idempotency guard skips documents whose stored offset already matches the target offset (safe replay after partial failure).
- **Memory pressure reduction**:
  - `FetchSourceContext` is used throughout to exclude unnecessary fields (`yaml`, full document bodies) from GET requests, including existence checks (`setFetchSource(false)`), space validations (`space.name` only), integration lookups, and ISM registration checks.
  - `ContentIndex.flush()` is called periodically (every 10 bulk batches) during both snapshot indexing and incremental updates to commit the translog and free heap held by small segments.
  - IOC hash computation switched from `StringBuilder` concatenation to streaming `MessageDigest`, avoiding large intermediate strings.
  - `deepCopy()` calls removed from `Cve` and `Ioc` model processing; fields are mutated in place.
  - Delete operations now use `WriteRequest.RefreshPolicy.NONE` instead of `IMMEDIATE`.
- **Circuit breaker resilience**: `ContentIndex.update()` retries up to 3 times with exponential backoff when a `CircuitBreakingException` is caught.
[...]

PR #1488, opened 2026-08-27
Add mechanism to retry bulk writes after shed

### LINKED ISSUE #1851: ContentIndex bulk indexing failure during Central Components install
**Describe the bug**

During the `5.0.0` Central Components DTT run (clustered Manager + Indexer + Dashboard, 10 all-in-one nodes), the `ContentIndex` component reports a bulk indexing failure right after install, on one of the 10 platforms (`amazon-2023-arm64`). This causes the CC log-check test `test_00_setup_install::test_check_logs[Indexer]` to fail.

**Evidence**

```
[2026-08-26T13:39:18,894][WARN ][c.w.c.c.c.i.ContentIndex ] [node-4] Bulk indexing finished with failures: failure in bulk execution:
```

This is the full extent of the message captured by the log check (single-line WARNING/ERROR grep over `wazuh-cluster.log`); no further stack trace or per-document failure detail was captured downstream. Only `amazon-2023-arm64` (node-4) reported it, at install time, before any restart/stop/uninstall step.

**Expected behavior**

Either the bulk indexing operation does not fail during a normal fresh install, or, if partial failures are an expected transient during initial content load, we'd like to know so we can add this message to the DTT log ignore list.

**Host/Environment**

- Wazuh version: 5.0.0-latest-beta5
- Component: Wazuh indexer (Central components, clustered AIO)
- Install method: Packages
- Platform: Amazon Linux 2023, arm64

**Additional context**

Execution: https://jenkins-staging.qa.wazuh.info/job/5.0.0/job/deployability_testing/1557/

Jorgesnchz:
## Description

It has been detected that the content manager silently discards CTI documents whenever the indexer sheds bulk writes under heap pressure.

The indexer is configured to shed writes rather than exhaust the heap, so a rejected bulk is an expected transient outcome that the client is responsible for re-submitting. `ContentIndex.executeBulk` never did, it logged a warning and moved on, and `SnapshotServiceImpl` then advanced the CTI consumer offset past the gap, so later incremental syncs skipped the missing documents

This PR solves this by retrying only the shed items with exponential backoff on the thread pool, counting anything finally given up on, and refusing to advance the consumer offset when documents were dropped, so the next sync re-runs instead of silently skipping the gap

Closes https://github.com/wazuh/wazuh-indexer/issues/1851

## Proposed Changes

- `ContentIndex.java`: Add retry what the cluster sheds

- `SnapshotServiceImpl.java`: Stop moving on over a gap

- `Constants.java`: Add new constants

### Results and Evidence

Reproduced on a 2-node cluster

I was not able to reproduce the DTT trigger, since it was real heap pressure. So I changed to `use_real_memory: false` and `indices.breaker.total.limit: 1kb`

I got the same error:
``` console
DTT   [0]: index [.wazuh-threatintel-vulnerabilities-a], id [CVE-2024-7295], message [
        RemoteTransportException[[node-9][172.31.43.44:9300][indices:data/write/bulk[s]]];
        nested: CircuitBreakingException[[parent] Data too large, ... in_flight_requests=31491564/30mb]];]

REPRO [1]: index [wazuh-threatintel-vulnerabilities-a],  id [CVE-2024-10001], message [
        RemoteTransportException[[node_t1][127.0.0.1:36021][indices:data/write/bulk[s]]];
        nested: CircuitBreakingException[[parent] Data too large, ... in_flight_requests=44994/43.9kb]];]
```

**After the changes**

Test 1: documents are recovered
``` console
[VALIDATION] parent circuit breaker limit set to [1kb]
[WARN ][c.w.c.c.c.i.ContentIndex ] [node_t1] Bulk indexing shed 33 document(s) under load; retry 1/3 in 1000ms.
[VALIDATION] parent circuit breaker limit set to [70%]
[VALIDATION][post-fix] 50 documents submitted, 50 indexed, 0 dropped
```

33 shed, all 33 recovered by the first retry. **50 / 50 indexed, 0 dropped.**

Test 2: pressure never clears (fails to recover)
``` console
[WARN ][c.w.c.c.c.i.ContentIndex ] [node_t2] Bulk indexing shed 50 document(s) under load; retry 1/3 in 1000ms.
[WARN ][c.w.c.c.c.i.ContentIndex ] [node_t2] Bulk indexing shed 50 document(s) under load; retry 2/3 in 2000ms.
[WARN ][c.w.c.c.c.i.ContentIndex ] [node_t2] Bulk indexing shed 50 document(s) under load; retry 3/3 in 4000ms.
[ERROR][c.w.c.c.c.i.ContentIndex ] [node_t2] Bulk indexing dropped 50 document(s) after 3 retries.
  Last failure: RemoteTransportException[[node_t1][127.0.0.1:34305][indices:data/write/bulk[s]]];
  nested: CircuitBreakingException[[parent] Data too large, data for [indices:data/write/bulk[s]]
  would be [71826/70.1kb], which is larger than the limit of [1024/1kb],
  usages [request=0/0b, fielddata=0/0b, in_flight_requests=71826/70.1kb]];
[VALIDATION][post-fix] pressure sustained: 50 documents submitted, 0 indexed, 50 dropped
```

Reproduce with:

```bash
./gradlew :wazuh-indexer-content-manager:test --tests "*ContentIndexBulkRetryTests*" -Dtests.logger.level=INFO
```

### Artifacts Affected

- Content Manager

### Configuration Changes

NA

### Documentation Updates

NA

### Tests Introduced

-`ContentIndexBulkRetryTests`: live 2-node `OpenSearchIntegTestCase` cluster, real circuit breaker, real bulk indexing:

| Test | Scope |
|---|---|
| `testCircuitBreakerShedsBulkWrites` | The cluster really sheds; captures the failure as the content manager receives it |
| `testFireAndForgetLosesShedDocuments` | Pre-fix baseline — the original listener, replicated verbatim, loses documents |
| `testExecuteBulkRecoversShedDocuments` | The fix recovers every shed document once pressure clears |
| `testExecuteBulkCountsDocumentsDroppedWhenPressureNeverClears` | Retry budget exhausts; every unlanded document is counted |

- `ContentIndexTests`: Add new tests that shed items retried

-`SnapshotServiceImplTests`: New test `testInitialize_DoesNotAdvanceOffsetWhenDocumentsWereDropped` drives `initialize` with a content index reporting `9332` dropped documents and asserts the consumer offset is never written.

#### Two things reviewers should know

> [!NOTE]
> On a single node I was not able to reproduce the error

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/index/ContentIndex.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/index/ContentIndex.java
@@ -43,13 +43,17 @@ import org.opensearch.action.index.IndexResponse;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.support.WriteRequest;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.breaker.CircuitBreakingException;
+import org.opensearch.core.concurrency.OpenSearchRejectedExecutionException;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.search.SearchHit;
 import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.search.fetch.subphase.FetchSourceContext;
+import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.client.Client;
 
 import java.io.IOException;
@@ -60,6 +64,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.wazuh.contentmanager.cti.catalog.model.*;
 import com.wazuh.contentmanager.cti.catalog.utils.JsonPatch;
@@ -84,6 +89,10 @@ public class ContentIndex {
     private static final long UPDATE_INITIAL_BACKOFF_MS = 1000;
     private static final long UPDATE_MAX_BACKOFF_MS = 30_000;
 
+    private static final int MAX_BULK_RETRIES = 3;
+    private static final long BULK_INITIAL_BACKOFF_MS = 1000;
+    private static final long BULK_MAX_BACKOFF_MS = 30_000;
+
     /** Maximum number of UPDATE offsets to batch into a single MultiGet + BulkRequest. */
     public static final int UPDATE_SUB_BATCH_SIZE = 50;
 
@@ -93,6 +102,14 @@ public class ContentIndex {
     private final Client client;
     private final PluginSettings pluginSettings;
     private final Semaphore semaphore;
+
+    /**
+     * Documents this instance failed to index and gave up on, either because the failure was not
+     * retryable or because the retry budget was exhausted. Callers loading a full snapshot must check
+     * this before advancing the consumer offset, otherwise the dropped documents are never
+     * backfilled.
+     */
+    private final AtomicLong droppedDocuments = new AtomicLong();
 
     /**
      * The public alias name (e.g., {@code "wazuh-threatintel-rules"}). All read operations use this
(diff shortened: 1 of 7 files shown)


ferooo23:
Testing this on a 2 node cluster: a snapshot whose indexing dropped 370090 of 380080 documents was still promoted to stable, and the rollback then loaded from that same file.

I think there is a bug here. `SnapshotServiceImpl` promotes before checking `getDroppedDocuments()`, so the same method refuses to advance `local_offset` over the load and still makes that zip the trusted fallback.

Still need to complete full validation.

ferooo23:
# Validation 🔴

A snapshot that could not be indexed in full is still promoted to stable, and the rollback then loads from that same snapshot. The safety net ends up being a file the system has just proven it cannot load.

Validated on a 2 node cluster.

## What happens

`SnapshotServiceImpl` promotes the downloaded snapshot before checking whether its indexing dropped documents:

| Path | Promotion | `dropped` check |
| --- | --- | --- |
| Remote (`initialize`) | `:214` | `:222-226` |
| Local (rollback) | `:461` | `:473-477` |

So the same method decides the load was bad enough not to advance `local_offset`, and good enough to make that zip the trusted fallback. Both cannot be true.

It is not a race. `success` is only set to `false` by an exception, but bulk failures are handled inside the listener and accumulated into `droppedDocuments` (`ContentIndex:798-801`), so dropped documents can never mark the load as failed. The javadoc of `promoteToStable` (`:525`) already states the precondition both call sites break: *"the path of the successfully-indexed snapshot"*.

## Where it comes from

`promoteToStable` predates this work: it was introduced in #1391. At that point there was no drop accounting, so there was nothing to contradict.

The `dropped > 0` check is new in this PR, and it was placed **after** the existing promotion instead of before it. That is where the contradiction appears.

## Evidence

No fault injection was needed. The drops came from the alias squatting bug, which leaves a content index without the plugin's mapping, saturated at 1000 fields:

```
[ERROR][c.w.c.c.c.i.ContentIndex ] [node-1] Bulk indexing dropped 37 document(s) with non-retryable failures. Last failure: java.lang.IllegalArgumentException: Limit of total fields [1000] has been exceeded

[ERROR][c.w.c.c.c.i.ContentIndex ] [node-1] [wazuh-threatintel-kvdbs] clear failed: InvalidAliasNameException[Invalid alias name [wazuh-threatintel-kvdbs], an index exists with the same name as the alias]
```

That part should be fixed by wazuh/wazuh-indexer-plugins#1486. It is only the trigger here: what follows is what the content manager does whenever documents are dropped, whatever the cause.

There was no `ruleset.stable.zip` in the snapshots directory before the run, so its appearance is the evidence:

```
07:55        no ruleset.stable.zip in the snapshots directory
07:56:27     Starting snapshot download [...beta5-t1-ruleset-5/1007_...zip]
07:56:32,461 Bulk indexing dropped 37 document(s) with non-retryable failures.
07:56:32,466 Snapshot indexing for consumer [cti:catalog:consumer:ruleset] dropped 37 document(s);
             local_offset was not advanced so the next sync retries the snapshot.
07:56:32,474 Remote snapshot failed. Rolling back from stable snapshot [.../ruleset.stable.zip]
07:56:56,389 Bulk indexing dropped 37 document(s)      <- the rollback, same failure
07:56:56,391 Snapshot indexing ... dropped 37 document(s); local_offset was not advanced
```

```
-rw-r--r--  554883  Aug 31 07:56  ruleset.stable.zip
```

The file did not exist before the run. It carries the timestamp of the run that declared itself incomplete, and the rollback consumed it 8 ms after the load was declared incomplete. Without the defect `stableExists` would have been `false`, there would have been no rollback, and the failure would have been clean instead of 24 seconds spent re-indexing a zip already proven unloadable. **The failed run manufactures its own rollback source, and that source is useless.**

The consumer ends up here, and the next cycle repeats the whole sequence:

```json
{ "status": "ready", "local_offset": 0, "remote_offset": 1007 }
```

## What "stable" means here

The zip is not corrupted. It downloads whole or not at all, and what is lost happens in transit from the zip to the index: the index ends up incomplete, the snapshot does not. So "stable" is not a claim about the file, it is a claim about the load:

> this zip was indexed in full at least once, so it works as a safety net

That claim is what goes unverified. A complete file is not the same as a loadable file.

With a transient cause the promotion is harmless, since the zip is loadable and the rollback works. The damage appears when the drops come from the data or from the target index, as above: the promotion overwrites a stable that was proven good with one that fails identically on every rollback.

## Proposed changes

Apply the rule this PR already applies to the offset: refuse to promote a load that dropped documents.

<details>
<summary>The change, in both paths</summary>

**Remote path** (`SnapshotServiceImpl.initialize(RemoteConsumer)`). Compute `dropped` right after the pending bulks settle, and return before the promotion block:

```java
if (success) {
    if (startMs != 0) { log.debug(Constants.D_LOG_SNAPSHOT_ELAPSED, ...); }

    // A partial load must not become the trusted fallback: "stable" means the snapshot
    // was indexed in full at least once. Same rule that keeps local_offset where it is.
    long dropped = this.getDroppedDocuments();
    if (dropped > 0) {
        log.error(Constants.E_LOG_SNAPSHOT_INDEXING_INCOMPLETE, this.consumerType, dropped);
        this.cleanup(snapshotZip);
        return false;
    }

    if (this.stablePath != null) {
        if (!promoteToStable(snapshotZip, this.stablePath)) {
            this.cleanup(snapshotZip);
        }
    } else {
        this.cleanup(snapshotZip);
    }
    return this.updateLocalOffset(consumer.getSnapshotOffset());
}
```

**Local path.** Move the same check ahead of the promotion block at `:460`. The existing `!localZip.equals(this.stablePath)` guard there already covers the rollback case, but not the packaged snapshot, which would still be promoted after a partial load.

**Resulting behaviour** on the run above: no `ruleset.stable.zip` is created, `stableExists` is `false`, there is no rollback from it, and the sync fails fast instead of re-indexing a zip already known to be unloadable. A previously good stable is preserved, which is the actual protection the three-tier fallback is meant to provide.

Worth adding a test that a load with `getDroppedDocuments() > 0` leaves `stablePath` untouched, in both paths.

</details>

ferooo23 (changes requested):
https://github.com/wazuh/wazuh-indexer-plugins/pull/1488#issuecomment-5477038867

ferooo23:
# Validation 🟢

Validated on a 2 node cluster, with the content-manager built from this branch.

`indexing_pressure.memory.limit` was lowered from its default `10%` to several values, to apply different levels of pressure and force the cluster to shed bulk writes. Each value sets how many of the 5 concurrent bulks fit, and therefore how much gets shed.

## Results

| Pressure | shed events | dropped docs | Docs indexed | `local_offset` | Stable promoted |
| --- | --- | --- | --- | --- | --- |
| `10%` (default) | 0 | 0 | 380080 | 905742 | yes |
| `22mb` | **198** | **0** | **380080** | **905742** | **yes** |
| `16mb` | 313 | 146 | 379934 | 0 | no |
| `1mb` | 1248 | all | 0 | 0 | no |
| field-limit rejection (#1476) | 0 | 37 | partial | 0 | no |

The five rows are consistent: the mechanism recovers what it can, is honest about what it loses, and the promotion is gated on documents actually being dropped, not on shedding having happened.

<details>
<summary>22mb: shed and fully recovered, stable promoted</summary>

The case the issue asks for. Four bulks fit, the fifth is rejected, so shedding is guaranteed while leaving room for the retries to land.

```
t=15s   docs=125088  shed=+50   dropped=+0
t=180s  docs=380080  shed=+198  dropped=+0   status: ready, local_offset: 905742
```

Retry chains that succeed, with the documented backoff and no third attempt:

```
09:35:25,382  Bulk indexing shed 92 document(s) under load; retry 1/3 in 1000ms.
09:35:26,385  Bulk indexing shed 92 document(s) under load; retry 2/3 in 2000ms.
09:35:30,261  Bulk indexing shed 999 document(s) under load; retry 1/3 in 1000ms.
09:35:31,266  Bulk indexing shed 999 document(s) under load; retry 2/3 in 2000ms.
```

Neither reaches `3/3` and neither is dropped. The full catalog lands and the stable is promoted:

```
-rw-r--r--  254903575  Sep  1 09:35  vulnerabilities.stable.zip
```

This is also what shows the promotion fix does not over-block: shedding occurred and the snapshot was still promoted, because only `getDroppedDocuments()` gates it.

</details>

<details>
<summary>16mb: shed, 146 documents lost, no promotion</summary>

Three bulks fit and the rest are shed. The retries recover nearly everything:

```
docs: 379934 of 380080     (146 lost)
local_offset: 0
snapshots: no vulnerabilities.stable.zip
```

Under sustained pressure the mechanism still gets 99.96% of the content in, and the residual loss is reported rather than hidden.

</details>

<details>
<summary>1mb: every bulk shed, retries exhausted, nothing promoted</summary>

Every bulk exceeds the limit on its own, so all three attempts fail. The rejection message carries the arithmetic:

```
rejected execution of coordinating operation [
  coordinating_and_primary_bytes=0,
  coordinating_operation_bytes=5296070,        <- one bulk
  max_coordinating_and_primary_bytes=1048576   <- the limit
]
```

The full backoff, traceable per bulk:

```
09:08:15,130  Bulk indexing shed 51 document(s) under load; retry 1/3 in 1000ms.
09:08:16,131  Bulk indexing shed 51 document(s) under load; retry 2/3 in 2000ms.
09:08:18,133  Bulk indexing shed 51 document(s) under load; retry 3/3 in 4000ms.
09:08:22,134  Bulk indexing dropped 51 document(s) after 3 retries.
```

1248 shed events, 0 documents indexed, `local_offset` held at 0, and no `vulnerabilities.stable.zip` created.

</details>

<details>
<summary>Field-limit rejection: non-retryable, dropped with no retry</summary>

No injection needed here: the alias squatting bug of #1476 leaves a content index without the plugin's mapping, saturated at 1000 fields, so 37 ruleset documents are rejected on every load. A mapping rejection is not retryable, so they are dropped immediately with no retry:

```
Bulk indexing dropped 37 document(s) with non-retryable failures.
Snapshot indexing for consumer [...] dropped 37 document(s); local_offset was not advanced
```

No stable created either. The promotion fix covers both drop causes, not just the retryable one.

</details>

## Before and after the fix, same trigger

The promotion defect was reproduced on this same cluster before the fix landed and re-run after:

```
07:20:34  dropped 37 · local_offset was not advanced
07:20:34  Rolling back from stable snapshot [.../ruleset.stable.zip]     <- promoted, then used
07:20:41  dropped 37 · local_offset was not advanced                     <- the rollback failing

07:23:04  dropped 37 · local_offset was not advanced                     <- no promotion
07:34:18  dropped 37 · local_offset was not advanced                     <- no rollback
```

The drop detection is unchanged, the bogus stable and the pointless rollback are gone.

ferooo23:
## On the DTT log check

The issue asks whether these failures are an expected transient, so the message can go on the ignore list. The answer differs per message, and one of the validation runs settles it: a load that finished **complete, with zero documents lost**, still emitted **198 WARN lines**.

| Message | Level | Transient? |
| --- | --- | --- |
| `Bulk indexing shed N document(s) under load; retry i/3 in Xms.` | WARN | **Yes.** Expected under load, the documents are recovered |
| `Bulk indexing dropped N document(s) after 3 retries.` | ERROR | No. Documents were lost |
| `Bulk indexing dropped N document(s) with non-retryable failures.` | ERROR | No. Documents were lost |

So the WARN belongs on the ignore list and the two ERRORs should keep failing the check, since they mean content is actually missing.

The message this issue reports, `Bulk indexing finished with failures: failure in bulk execution:`, no longer exists in production code. Any log expectation matching it can be dropped.

ferooo23 (approved):
LGTM!
Validation: https://github.com/wazuh/wazuh-indexer-plugins/pull/1488#issuecomment-5492335435

AlexRuiz7:
Component tests pass:
- https://github.com/wazuh/wazuh-indexer-plugins/actions/runs/33508675022/job/99863643662

AlexRuiz7:
New logs added to https://github.com/wazuh/wazuh-indexer-plugins/pull/1488

PR #1506, opened 2026-08-31
Retry sending resources to SAP after failure

### LINKED ISSUE #1487: A failed detector sync is abandoned permanently, so one transient error leaves the deployment with no detection
|---|---|---|---|---|
| 5.0.0 | wazuh-indexer-plugins (content-manager: `ConsumerRulesetService`, `AbstractConsumerService`) | Manager | Docker, `public.ecr.aws/wazuh-cicd/wazuh/*:5.0.0-latest` | Amazon Linux 2023 |

## Description

`ConsumerRulesetService.syncDetectors()` pushes the 21 detectors that carry the entire ruleset. If a push fails, the failures are collected into a list, logged as one `WARN`, and dropped. There is no retry, no backoff, and no state recorded that a retry is owed.

The next sync pass does not pick it up either, and that is what makes it permanent. `onSyncComplete(boolean isUpdated)` gates the whole Security Analytics block behind `if (isUpdated)`, which is true only on a pass that actually loaded a snapshot or applied an incremental update. Once content is local, every subsequent `CatalogSyncJob` pass returns `updated=false`, so `syncDetectors` is never called again. Restarting the indexer does not help, for the same reason.

A deployment that loses its detector push therefore has no detectors, and nothing in the product will ever try again. The only thing that would is a new content offset published upstream by CTI, which is unpredictable and may be days away.

The stack looks healthy meanwhile. Agents connect, events arrive, decoders normalize them into ECS, events are indexed, and the inventory, FIM and SCA state tables fill up. The Threat Hunting views are simply empty, which a user cannot distinguish from "nothing has happened yet".

## Configuration/Considerations

Stock single-node deployment, no configuration changes.

**The trigger is any transient failure during the single sync pass that loads content.** Two independent causes are known, and a fix aimed at only one leaves the other open:

- `cluster.blocks.create_index` being up when the push runs. This block is set by OpenSearch core, not by any Wazuh component: `DiskThresholdMonitor.handleClusterCreateIndexBlocks` applies it as a **persistent** setting when every node breaches the high disk watermark (default `90%`), and auto-releases it when they no longer do (`cluster.blocks.create_index.auto_release`, default `true`). A grep for that setting across the `wazuh` org returns nothing outside OpenSearch's own definition. Nothing here is a Wazuh setting and nothing about it needs changing.
- The `wazuh-events-v5-*` data streams not existing yet when the push runs. This needs no disk pressure at all: the trigger is component startup order.

**Detector creation and detector execution behave differently, and only the first is a defect.** Under the same cluster block, the scheduler that runs detectors logs a failure every minute and recovers on its own the moment the block clears. Measured: with the block still up, `Workflow run failed` repeated once a minute from 09:14:30; at 09:21:04 the lock index was created, at 09:22:03 there were 65 findings, including 64 rootcheck events that had been sitting unmatched in the events index. Nothing was restarted or re-pushed.

| | retries? | recovers when the condition clears? |
| --- | --- | --- |
| detector **creation** (`syncDetectors`) | no | **no, ever** |
| detector **execution** (`MonitorRunnerService` / `LockService`) | yes, every minute | yes, automatically |

## Steps to reproduce

**Producing the failure**, deterministically and with no disk pressure. Put the block up by hand immediately before the content load completes, then take it down again:

```bash
curl -sk -u admin:admin -X PUT 'https://<indexer>:9200/_cluster/settings' \
  -H 'Content-Type: application/json' -d '{"persistent":{"cluster.blocks.create_index":true}}'
# wait for "detectors could not be sent" in the indexer log
curl -sk -u admin:admin -X PUT 'https://<indexer>:9200/_cluster/settings' \
  -H 'Content-Type: application/json' -d '{"persistent":{"cluster.blocks.create_index":null}}'
```

**Confirming nothing repairs it**, with the block gone and the cluster fully healthy. This is the finding; everything else is one way of arriving at it:

```bash
# 0 for the life of the deployment
curl -sk -u admin:admin -X POST 'https://<indexer>:9200/_plugins/_security_analytics/detectors/_search' \
  -H 'Content-Type: application/json' -d '{"query":{"match_all":{}},"size":0}'

docker restart <indexer>
# still 0, and no new ConsumerRulesetService line is logged at all
docker logs <indexer> 2>&1 | grep -c "ConsumerRulesetService"
```

**As originally observed.** Fill the host above the 90% disk watermark and deploy. OpenSearch raises the block during content bootstrap and the push lands inside it. Not deterministic: of three consecutive deployments on the same host, one lost the push and two did not.

A deployment that comes up with 21 detectors is not evidence that this is fixed. Only the restart check is.

## Current behavior

Every findings data stream is empty on a 77-minute-old deployment, and not a single detector exists:

```
$ curl -X POST '_plugins/_security_analytics/detectors/_search' -d '{"query":{"match_all":{}},"size":0}'
{"hits":{"total":{"value":0,"relation":"eq"},"hits":[]}}

$ curl '_cat/indices/.opensearch-sap*'
.opensearch-sap-pre-packaged-rules-config 3187
.opensearch-sap-log-types-config           162
```

The plugin is installed and holds its pre-packaged rules and log types. What is missing is the detector configuration index, created lazily on the first detector write.

The push is one operation in three parts and only the third fails. Integrations and rules reach the plugin; the detectors, which require that configuration index, do not:

```
07:13:33 [INFO][ConsumerRulesetService] Sent 125 of 125 integrations to Security Analytics for space [standard].
07:13:38 [INFO][ConsumerRulesetService] Sent 290 of 290 rules to Security Analytics for space [standard].
07:13:38 [WARN][ConsumerRulesetService] 21 detectors could not be sent to Security Analytics for space [standard]: [...]
```

Those three lines are the entire history of the service for the life of the deployment, and the underlying error appears once:

```
07:13:38 [ERROR][o.o.s.u.SecurityAnalyticsException] Security Analytics error:
    at org.opensearch.securityanalytics.util.DetectorIndices.initDetectorIndex(DetectorIndices.java:59)
07:13:38 [ERROR][o.o.s.t.TransportIndexDetectorAction] exception: blocked by: [FORBIDDEN/10/cluster create-index blocked (api)];
07:13:38 [WARN ][ConsumerRulesetService] 21 detectors could not be sent: [azure, suricata, windows, github,
  wazuh-manager, wazuh-generic-1, o365, wazuh-it-hygiene, cloud-detection, docker, wazuh-fim, wazuh-rootcheck,
  apache-http, linux, wazuh-vd, gcp, auditd, wazuh-sca, aws, wazuh-ioc, wazuh-generic-2]
```

That list is the whole product. Attempted exactly once:

```
$ docker logs <indexer> | grep -c "detectors could not be sent"
1
```

### Permanence, measured

Verified on a purpose-built 5.0.0 stack, using the stack's own healthy push as the control.

On first boot, with nothing blocking it, all 21 detectors were created:

```
11:12:56 Sent 125 of 125 integrations to Security Analytics for space [standard].
11:13:01 Sent 290 of 290 rules to Security Analytics for space [standard].
11:13:05 Sent 21 of 21 detectors to Security Analytics for space [standard].
```

Deleting detectors through the API is refused, which removes the last manual escape:

```
DELETE /_plugins/_security_analytics/detectors/<id>
{"error":{"reason":"Standard detectors cannot be deleted by users."},"status":400}
```

The affected state was produced instead by dropping the configuration index, which is what a failed push leaves behind. The indexer was then restarted at 13:45:25 and was up 15 seconds later. Three minutes on:

```
detectors: 0
findings:  0
CatalogSyncJob passes since the restart:        1
ConsumerRulesetService lines since the restart: 0
```

Across the entire life of the container `ConsumerRulesetService` logged exactly three lines, all from the first boot. A sync pass ran after the restart and never reached the Security Analytics block, precisely as the `if (isUpdated)` gate predicts.

This reproduces the state a failed push leaves behind rather than the push failure itself, so it establishes the permanence, not the trigger.

### Everything upstream of the detector works, which is what makes this invisible

Injecting six sshd lines into a monitored file produced six correctly normalized events, indexed in the exact data stream the `linux` integration's detector declares as its source, and zero findings. The decoded event carries `wazuh.integration.name: linux`, `event.action: logged-in`, `event.category: [authentication, session]`, `user.name: root`; the loaded ruleset contains a rule selecting exactly that. Counters confirm nothing is dropped on the way in:

```
347  server.events.received
347  router.events.processed
  0  events.discarded
  0  indexer.events.dropped
```

Decoding, classification, routing and indexing are all fine. There is simply nothing subscribed to the events.

## Root cause

Three properties of `ConsumerRulesetService` in `wazuh-indexer-plugins` combine to make one failure permanent.

**1. The whole block is gated on the pass having changed something.**

```java
// ConsumerRulesetService.java:160
public void onSyncComplete(boolean isUpdated) {
    this.initializeSpaces();

    if (isUpdated) {                                    // :163
        ...
        if (PluginSettings.getInstance().getCreateDetectors()) {
            try {
                this.syncDetectors(integrationDocs);    // :189
            } catch (Exception e) {
                log.error(Constants.E_LOG_SAP_SYNC_FAILED, "detectors", e.getMessage(), e);
            }
        }
    }
}
```

`isUpdated` comes from `AbstractConsumerService.synchronize()`, which passes `SyncResult.updated()`. That is true only where a snapshot was swapped in or `performIncrementalUpdate` returned true. With content already local and no new remote offset it is `false`, and the block is skipped in full.

**2. A partial failure is logged once and not recorded.**

```java
// ConsumerRulesetService.java:520
if (!failed.isEmpty()) {
    log.warn(Constants.W_LOG_SAP_PARTIAL, failed.size(), "detectors", Space.STANDARD, failed);
}
```

That is the last statement in the method. No rethrow, no flag, no reschedule; `failed` goes out of scope with the method.

**3. The missing-source-index path abandons just as permanently.**

```java
// ConsumerRulesetService.java:433
List<String> missingIndices = this.missingSourceIndices(docs);
if (!missingIndices.isEmpty()) {
    log.error("Skipping detectors sync. Required source indices not found: {}",
            String.join(", ", missingIndices));
    return;
}
```

Same shape, same consequence, and no cluster block or full disk required.

**There is no other path that could repair it.** The only production call sites of `upsertDetectorAsync` are `ConsumerRulesetService.java:450` and `:486`, both inside `syncDetectors`; the remaining references are the interface, its implementation and `MockSecurityAnalyticsService`.

## Expected results

A deployment either has its detectors installed or says loudly that it does not.

- A detector push that fails is retried on subsequent sync passes until it succeeds, rather than abandoned on the one pass where `isUpdated` happened to be true.
- The same applies to the `missingSourceIndices` skip, which abandons just as permanently for an unrelated reason.
- A stack with zero detectors is a broken stack and says so: in the manager log, in the cluster health the UI shows, and in the Threat Hunting views, which today render an empty result set indistinguishable from a quiet environment.

## Impact

When it happens, the product's core function is off for the life of the deployment. No rule is ever evaluated, so no finding is ever produced, while agents connect, events are ingested and normalized correctly, and every other health signal is green.

There is no recovery path available to an operator. A restart does not help, because the sync pass that would re-push is a no-op once content is local. The detectors cannot be re-created by hand either, since the API refuses to delete standard detectors. The only thing that would trigger a fresh push is a new content offset published upstream by CTI.

Because the failure is a lost race, a deployment that boots healthy proves nothing about the next one on the same host.

## Tasks

- [x] Retry the detector sync until it succeeds. The smallest fix is a field on `ConsumerRulesetService` set when `failed` is non-empty **or** when the `missingSourceIndices` guard trips, OR'd into the `if (isUpdated)` condition in `onSyncComplete`, so the next `CatalogSyncJob` pass re-attempts. That covers both abandon paths and the cluster-block case at once, since the following pass runs after the block auto-releases.
- [x] Decide whether the `if (isUpdated)` gate is right for integrations and rules too. Detectors are the visible casualty, but the gate skips all three, so any partial failure in the other two is equally permanent.
- [ ] Raise the failure past a single `WARN` in the indexer log: the manager should log it, and the UI should show that detection is not running.
- [ ] Reconsider whether standard detectors should be undeletable through the API, which today leaves an affected deployment with no manual remedy at all.
- [ ] Treat `wazuh-indexer-plugins#1476` and `wazuh-indexer-plugins#1481` as sharing this one's **trigger, not its defect**. Both were observed contemporaneously with it, because the same cluster block refuses whatever index creation happens to be in flight. The question to ask of each separately is the one answered here: when the condition clears, does that component recover on its own? Detector execution does; detector creation does not.
- [ ] Add a check asserting the expected detectors exist after deployment, since every other health signal is green while this is broken. Assert after a restart, not just after boot, or the check will pass on a deployment that is permanently broken.

MauriFont:
## Description

The Security Analytics content sync (Integrations, Rules, Detectors) previously ran only when the catalog had new remote content to apply, so a transient Security Analytics failure on one sub-phase was never retried until unrelated catalog content changed — potentially never.

Closes #1487 

## Proposed Changes

This adds a small persisted state, pending_sync_phases, to the .wazuh-cti-consumers document that tracks which of the three sub-phases (integrations, rules, detectors) still owe a retry:

- Each sub-phase reports success or failure independently. A phase that fails is recorded as pending; a phase that succeeds is cleared.
- On every sync pass, a phase runs if either the catalog has new content or it was left pending from a previous pass — so a failure gets retried on the very next scheduled sync, not on the next unrelated content update.
- Integrations and Detectors share the same underlying read of the integrations index; that read is done once per pass and its outcome (not just its result being empty) determines whether Detectors can be evaluated, so a read failure keeps Detectors pending rather than mistaking "nothing read" for "nothing to sync."
- A detectors phase left pending while detector creation is disabled via settings is dropped instead of being retried forever, since it can never succeed in that configuration.
- The consumer document is only rewritten when the set of pending phases actually changes, so a healthy pass or a phase that keeps failing the same way doesn't pay a write on every sync.

Example document:

```
{
  "_index" : ".wazuh-cti-consumers",
  "_id" : "cti:catalog:consumer:ruleset",
  "_version" : 14,
  "_seq_no" : 225,
  "_primary_term" : 22,
  "found" : true,
  "_source" : {
    "name" : "public-ruleset-5",
    "context" : "beta5-t1-ruleset-5",
    "type" : "cti:catalog:consumer:ruleset",
    "resource" : "...",
    "is_public" : true,
    "status" : "ready",
    "local_offset" : 1003,
    "remote_offset" : 1003,
    "pending_sync_phases" : [      <--- new
      "integrations",
      "rules",
      "detectors"
    ]
  }
}

```
### Results and Evidence

First run with detectors failure
Captura desde 2026-08-31 16-58-52

Next run the detectors are sent successfully
Captura desde 2026-08-31 16-59-06

### Artifacts Affected

- Executables: `wazuh-indexer-content-manager` plugin (Java changes)
- Default configuration files: `consumers-mapping.json` (new `pending_sync_phases` keyword field)

### Configuration Changes

N/A

### Documentation Updates

- docs/ref/modules/content-manager/architecture.md
- docs/ref/modules/content-manager/troubleshooting.md
- docs/ref/modules/content-manager/index.md

### `LocalConsumerTests.java` — 4 new tests

Covers the `LocalConsumer` model and the new `pendingSyncPhases` field:

| Test | What it verifies |
|---|---|
| `testPendingSyncPhasesDefaultsEmpty` | The constructor without that parameter initializes it to empty, not `null`. |
| `testToXContentIncludesPendingSyncPhases` | Serializes correctly to JSON. |
| `testPendingSyncPhasesDeserializesFromJson` | Deserialization round-trip. |
| `testLegacyDocumentWithoutPendingSyncPhasesDefaultsEmpty` | A legacy document without the field (pre-migration) deserializes to an empty list instead of failing. |

### `ConsumerRulesetServiceTests.java` — 10 new tests

All exercise `onSyncComplete`/`synchronize` on `ConsumerRulesetService`:

| Test | What it verifies |
|---|---|
| `testOnSyncComplete_integrationsPartialFailure_setsOnlyIntegrationsPending` | An isolated integrations failure marks only that phase pending, without touching rules/detectors. |
| `testOnSyncComplete_rulesPartialFailure_setsOnlyRulesPending` | Symmetric case for rules. |
| `testOnSyncComplete_missingSourceIndices_setsDetectorsPending` | Missing source indices for detectors leaves that phase pending, without invoking `upsertDetectorAsync`. |
| `testOnSyncComplete_notUpdatedButDetectorsPending_onlyRetriesDetectors` | The core regression case: with `isUpdated=false` but `detectors` pending, only that phase is retried (integrations is read but not re-sent, rules is never touched). |
| `testOnSyncComplete_fullSuccess_clearsPendingPhases` | If all three pending phases succeed on retry, the persisted list is cleared. |
| `testOnSyncComplete_notUpdatedAndNothingPending_isNoOp` | With no new content and nothing pending, there are no reads or writes — a true no-op. |
| `testSynchronize_pendingPhaseSurvivesTrailingStatusWrite` | The trailing `status=READY` write at the end of `synchronize()` does not clobber `pending_sync_phases`. |
| `testOnSyncComplete_integrationsReadFailure_keepsDetectorsPending` | A failed integrations read must not be mistaken for "nothing to sync": `detectors` stays pending, and there's no write (nothing changed). |
| `testOnSyncComplete_detectorsDisabled_dropsUnreachableDetectorsPhase` | With `create_detectors=false`, a pending `detectors` phase is dropped instead of being retried forever. |
| `testOnSyncComplete_pendingPhasesUnchanged_skipsConsumerWrite` | If the set of pending phases doesn't change between passes, the consumer document is not rewritten. |

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,7 +185,7 @@
 - [BUG] Fix issues in workflows [(#1449)](https://github.com/wazuh/wazuh-indexer/issues/1449)
 - Failed to save/update ManagedIndexMetaData [(#1828)](https://github.com/wazuh/wazuh-indexer/issues/1828)
 - [BUG] `wazuh-threatintel-enrichments-a` is never created, so the CTI writer auto-creates a dynamically mapped index under the alias name and the home overview's Threat catalog fails [(#1476)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1476)
-- [BUG] SAP sync not retrying after failure [(#1487)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1487)
+- [BUG] Retry sending resources to SAP after failure [(#1487)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1487)
 
 ## Prior versions
 - []()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -182,6 +182,7 @@
 - [BUG] Fix issues in workflows [(#1449)](https://github.com/wazuh/wazuh-indexer/issues/1449)
 - Failed to save/update ManagedIndexMetaData [(#1828)](https://github.com/wazuh/wazuh-indexer/issues/1828)
 - [BUG] `wazuh-threatintel-enrichments-a` is never created, so the CTI writer auto-creates a dynamically mapped index under the alias name and the home overview's Threat catalog fails [(#1476)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1476)
+- [BUG] SAP sync not retrying after failure [(#1487)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1487)
 
 ## Prior versions
 - []()
--- a/docs/ref/modules/content-manager/architecture.md
+++ b/docs/ref/modules/content-manager/architecture.md
@@ -255,7 +255,8 @@ The `.wazuh-cti-consumers` index stores one document per consumer type:
     "is_public": true,
     "status": "ready",
     "local_offset": 3932,
-    "remote_offset": 3932
+    "remote_offset": 3932,
+    "pending_sync_phases": []
   }
 }
 ```
@@ -269,3 +270,5 @@ The `status` field reflects the consumer's synchronization lifecycle:
 | `failed` | The previous sync cycle was interrupted by an unexpected exception. |
 
 The status is set to `running` at the very start of a sync cycle and transitions to `ready` after all post-sync work finishes — including hash recalculation, Security Analytics Plugin synchronization, and Engine IoC notification — or to `failed` if an unexpected exception interrupts the cycle. The job scheduler logs the failure and retries on the next scheduled run regardless of the consumer's status.
+
+For the ruleset consumer, `pending_sync_phases` lists which of `integrations`, `rules`, and `detectors` failed to sync to Security Analytics on the last pass and are retried on the next scheduled sync, even without new CTI content; an empty array means all sub-phases are in sync.
--- a/docs/ref/modules/content-manager/index.md
+++ b/docs/ref/modules/content-manager/index.md
@@ -105,7 +105,7 @@ The Content Manager uses the following system indices:
 
 | Index                                | Description                                                                                          |
 | -------------------------------------- | ----------------------------------------------------------------------------------------------------- |
-| `.wazuh-cti-consumers`                | Synchronization state for each CTI consumer type (`type`, `resource`, `is_public`, offsets, status)   |
+| `.wazuh-cti-consumers`                | Synchronization state for each CTI consumer type (`type`, `resource`, `is_public`, offsets, status, pending sync phases) |
 | `.wazuh-internal-state`               | Persisted CTI access token (hidden, single document)                                                  |
 | `wazuh-threatintel-rules`             | Detection rules (both CTI-synced and user-generated, across all spaces)                               |
 | `wazuh-threatintel-decoders`          | Log decoders                                                                                           |
--- a/docs/ref/modules/content-manager/troubleshooting.md
+++ b/docs/ref/modules/content-manager/troubleshooting.md
@@ -148,7 +148,8 @@ Example output:
           "is_public": true,
           "status": "ready",
           "local_offset": 3932,
-          "remote_offset": 3932
+          "remote_offset": 3932,
+          "pending_sync_phases": []
         }
       }
     ]
@@ -162,6 +163,7 @@ Example output:
 - `local_offset == remote_offset`: Content is up-to-date.
 - `local_offset < remote_offset`: Content needs updating.
 - `local_offset == 0`: Content has never been synced (snapshot required).
+- `pending_sync_phases`: for the ruleset consumer, lists which of `integrations`, `rules`, and `detectors` failed to sync to Security Analytics and are retried on the next scheduled sync; empty means all sub-phases are in sync.
 
 ### Check sync job
 
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/model/LocalConsumer.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/model/LocalConsumer.java
@@ -26,6 +26,8 @@ import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.core.xcontent.XContentBuilder;
 
 import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
 
 /**
  * Data Transfer Object representing the local state of a CTI Catalog Consumer.
@@ -82,6 +84,9 @@ public class LocalConsumer extends AbstractConsumer implements ToXContent {
 
     @JsonProperty("is_public")
     private boolean isPublic;
+
+    @JsonProperty("pending_sync_phases")
+    private List<String> pendingSyncPhases = Collections.emptyList();
 
     /** Default constructor. */
     public LocalConsumer() {
@@ -171,6 +176,41 @@ public class LocalConsumer extends AbstractConsumer implements ToXContent {
     }
 
     /**
+     * Constructs a LocalConsumer with full state details including explicit status and pending SAP
+     * sync phases.
+     *
+     * @param context The context identifier.
+     * @param name The consumer name.
+     * @param type The consumer type identifier.
+     * @param resource The full CTI consumer URL.
+     * @param isPublic Whether the consumer is public.
+     * @param status The current synchronization status.
+     * @param localOffset The current offset processed locally.
+     * @param remoteOffset The last known offset available remotely.
+     * @param pendingSyncPhases Names of SAP sync sub-phases still owed a retry.
+     */
+    public LocalConsumer(
+            String context,
+            String name,
+            String type,
+            String resource,
+            boolean isPublic,
+            Status status,
+            long localOffset,
+            long remoteOffset,
+            List<String> pendingSyncPhases) {
+        this.context = context;
+        this.name = name;
+        this.type = type;
+        this.resource = resource;
+        this.isPublic = isPublic;
+        this.status = status;
+        this.localOffset = localOffset;
+        this.remoteOffset = remoteOffset;
+        this.pendingSyncPhases = pendingSyncPhases != null ? pendingSyncPhases : Collections.emptyList();
+    }
+
+    /**
      * Gets the current synchronization status of the consumer.
      *
      * @return The {@link Status} indicating whether the consumer is ready, running, or failed.
@@ -210,6 +250,16 @@ public class LocalConsumer extends AbstractConsumer implements ToXContent {
     /** Returns true if the consumer is public. */
     public boolean isPublic() {
         return this.isPublic;
+    }
+
+    /**
+     * Gets the names of SAP sync sub-phases still owed a retry.
+     *
+     * @return The pending phase names (e.g. "integrations", "rules", "detectors"), or an empty list
+     *     if none are pending.
+     */
+    public List<String> getPendingSyncPhases() {
+        return this.pendingSyncPhases;
     }
 
     @Override
@@ -268,6 +318,7 @@ public class LocalConsumer extends AbstractConsumer implements ToXContent {
                 .field("status", this.status != null ? this.status.toString() : Status.READY.toString())
                 .field("local_offset", this.localOffset)
                 .field("remote_offset", this.remoteOffset)
+                .field("pending_sync_phases", this.pendingSyncPhases)
                 .endObject();
 
         return builder;
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/model/LocalConsumer.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/model/LocalConsumer.java
@@ -207,7 +207,8 @@ public class LocalConsumer extends AbstractConsumer implements ToXContent {
         this.status = status;
         this.localOffset = localOffset;
         this.remoteOffset = remoteOffset;
-        this.pendingSyncPhases = pendingSyncPhases != null ? pendingSyncPhases : Collections.emptyList();
+        this.pendingSyncPhases =
+                pendingSyncPhases != null ? pendingSyncPhases : Collections.emptyList();
     }
 
     /**
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/AbstractConsumerService.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/AbstractConsumerService.java
@@ -253,11 +253,71 @@ public abstract class AbstractConsumerService {
                             current.isPublic(),
                             status,
                             current.getLocalOffset(),
-                            current.getRemoteOffset());
+                            current.getRemoteOffset(),
+                            current.getPendingSyncPhases());
             this.consumersIndex.setConsumer(updated);
             log.debug(Constants.D_LOG_CONSUMER_STATUS_SET, consumerType, status);
         } catch (Exception e) {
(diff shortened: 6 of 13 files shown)


ferooo23:
# Validation 🟢

Validated on a single-node AIO installed from today's nightly packages, with the content-manager plugin built from this branch.

To fail one phase at a time, writes were blocked on the index that phase writes to. The issue uses `cluster.blocks.create_index`, which fails everything at once; blocking a single index lets one phase fail while the others go through.

In both cases the content load itself finished, so `local_offset` and `remote_offset` matched and `isUpdated` was `false`. That is the condition this PR is about: without it the whole Security Analytics block is skipped, and the phase that failed is never sent again.

The retry would normally arrive on the scheduled pass, an hour later by default (`catalog.sync_interval`). Rather than wait for it, each retry here was triggered with `POST _plugins/_content_manager/update`, which goes through the same code. What shows the retry went through the pending marker and not through new content is that `local_offset` does not move between the failure and the recovery.

## 1. Rules

Blocked `.opensearch-sap-pre-packaged-rules-config`:

```
14:05:35  Sent 126 of 126 integrations to Security Analytics for space [standard].
14:05:36  328 rules could not be sent to Security Analytics for space [standard]: [...]
14:05:36  22 detectors could not be sent to Security Analytics for space [standard]: [...]
14:05:36  Security Analytics content sync degraded for consumer [cti:catalog:consumer:ruleset]:
          phase(s) [rules, detectors] still pending; will retry on the next scheduled sync pass.
```

Recorded as pending, with the content load itself complete:

```json
{ "status": "ready", "local_offset": 1029, "remote_offset": 1029,
  "pending_sync_phases": ["rules", "detectors"] }
```

Block removed and a sync triggered with no new content:

```
14:06:31  Sent 328 of 328 rules to Security Analytics for space [standard].
14:06:46  Sent 22 of 22 detectors to Security Analytics for space [standard].
```

```json
{ "status": "ready", "local_offset": 1029, "pending_sync_phases": [] }
```

Detectors failed along with the rules, which is expected: a detector attaches rules by id, so with the rules unwritten there is nothing to attach. Both are recorded and both recover.

## 2. Integrations

Blocked `.opensearch-sap-log-types-config`. Only the blocked phase is recorded:

```
14:35:36  126 integrations could not be sent to Security Analytics for space [standard]: [...]
14:35:38  Sent 328 of 328 rules to Security Analytics for space [standard].
14:35:54  Sent 22 of 22 detectors to Security Analytics for space [standard].
14:35:54  Security Analytics content sync degraded for consumer [cti:catalog:consumer:ruleset]:
          phase(s) [integrations] still pending; will retry on the next scheduled sync pass.
```

```json
{ "status": "ready", "local_offset": 1033, "pending_sync_phases": ["integrations"] }
```

Rules and detectors have their own `Sent` lines, so they ran and completed instead of being skipped.

Detectors did not fail here because the log types that the integrations phase writes were already in the index from earlier runs, so `detector_type` still resolved. On a fresh install they would not be there yet, and an integrations failure would probably take the detectors with it.

Retried with no new content and cleared:

```json
{ "status": "ready", "local_offset": 1033, "pending_sync_phases": [] }
```

# Suggestion

One small thing I ran into while reviewing, just flagging it in case you want it in.

`toXContent` writes the field even when it is empty, so every consumer document now carries an empty array, `iocs` and `vulnerabilities` included,:

```json
{ "_id": "cti:catalog:consumer:ruleset",         "pending_sync_phases": [] }
{ "_id": "cti:catalog:consumer:iocs",            "pending_sync_phases": [] }
{ "_id": "cti:catalog:consumer:vulnerabilities", "pending_sync_phases": [] }
```

Could be changed for cleanliness, guarding it with `if (!this.pendingSyncPhases.isEmpty())` so the healthy state is an absent field.

ferooo23 (dismissed):
LGTM!
Validation: https://github.com/wazuh/wazuh-indexer-plugins/pull/1506#issuecomment-5496315600

CHANGELOG.md
AlexRuiz7, line 1: Use the PR title as change entry.

MauriFont: Done

AlexRuiz7 (dismissed):

ferooo23 (dismissed):

AlexRuiz7 (approved):

PR #1517, opened 2026-09-02
Explicitly send Accept-Encoding: gzip, br on all CTI/GIP requests

MauriFont:
## Description

Every request the XDR Indexer (content-manager plugin) makes against the CTI service must include the compression acceptance header `Accept-Encoding: gzip`, so that response payloads are transferred compressed instead of as plain text.

HttpClient5 already negotiates gzip transparently by default (`Accept-Encoding: gzip, x-gzip, deflate`, auto-decompressed), so this doesn't add a capability the client didn't already have. The point is to make the exact header value explicit **in code**, on every request, instead of leaving it to the library's implicit default: HttpClient5's own default value includes `x-gzip` and `deflate`, which CTI doesn't need, and relying on a default means the actual wire behavior isn't visible anywhere in this codebase.

## Proposed Changes

- Added an explicit `Accept-Encoding: gzip` header to every outbound HTTP request the content-manager plugin makes to CTI:
  - Catalog `ApiClient` (`getConsumer`, `getChanges`, `getReleaseUpdates`)
  - `SnapshotClient` (snapshot downloads)
  - Console `ApiClient` (token exchange, resource token, plans, environment lookup, catalog plans)
  - `TelemetryClient` (`/ping`)
- The header is set **per request** rather than as a client-level default header. HttpClient5's exec chain runs `ChainElement.COMPRESS` before `ChainElement.PROTOCOL` (where default headers are applied), so its content-compression handling adds its own `Accept-Encoding` first if none is present yet, and the default-header step then skips a header name that's already there — a default header would never actually be inserted. Setting it directly on each request is what reaches the wire as `gzip`.
- Centralized the header value in a new `Constants.ACCEPT_ENCODING_GZIP` constant instead of repeating the string literal `"gzip"` at each of the 12 call sites.
- Added debug-level logging in `HttpResponseCallback` reporting the request's `Accept-Encoding`, the response's `Content-Encoding`, the compressed wire size, and the decompressed body size, to verify compression is negotiated correctly. (Useful for testing this PR; can be removed.)

### Results and Evidence

Captura desde 2026-09-04 16-16-57

imagen

### Artifacts Affected

- wazuh-indexer-plugins -> content-manager

### Configuration Changes

N/A

### Documentation Updates

- docs/ref/modules/content-manager/architecture.md
- docs/ref/modules/content-manager/configuration.md
- docs/dev/plugins/content-manager.md

### Tests Introduced

N/A — no new tests added. The header value change is covered indirectly by the existing `ApiClient`/`SnapshotClient`/`TelemetryClient` test suites, but no dedicated test asserts the `Accept-Encoding` header value itself.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/console/client/TelemetryClient.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/console/client/TelemetryClient.java
@@ -20,10 +20,12 @@ import org.apache.hc.client5.http.async.methods.SimpleHttpRequest;
 import org.apache.hc.client5.http.async.methods.SimpleHttpResponse;
 import org.apache.hc.client5.http.async.methods.SimpleRequestBuilder;
 import org.apache.hc.core5.concurrent.FutureCallback;
+import org.apache.hc.core5.http.HttpHeaders;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import com.wazuh.contentmanager.settings.PluginSettings;
+import com.wazuh.contentmanager.utils.Constants;
 
 /**
  * Client for handling telemetry-related requests to the Wazuh CTI API. This class extends the base
@@ -52,7 +54,7 @@ public class TelemetryClient extends ApiClient {
                             .addHeader("wazuh-uid", uuid)
                             .addHeader("wazuh-tag", "v" + version)
                             .addHeader("Accept", "application/json")
-                            .addHeader("Accept-Encoding", "gzip, br")
+                            .addHeader(HttpHeaders.ACCEPT_ENCODING, Constants.ACCEPT_ENCODING_GZIP)
                             .build();
 
             log.debug("Sending telemetry ping to: {}", pingUrl);
--- a/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/console/client/TelemetryClient.java
+++ b/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/console/client/TelemetryClient.java
@@ -52,6 +52,7 @@ public class TelemetryClient extends ApiClient {
                             .addHeader("wazuh-uid", uuid)
                             .addHeader("wazuh-tag", "v" + version)
                             .addHeader("Accept", "application/json")
+                            .addHeader("Accept-Encoding", "gzip, br")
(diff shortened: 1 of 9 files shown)


plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/console/client/TelemetryClient.java
```diff
@@ -52,6 +52,7 @@ public void ping(String uuid, String version) {
                             .addHeader("wazuh-uid", uuid)
                             .addHeader("wazuh-tag", "v" + version)
                             .addHeader("Accept", "application/json")
+                            .addHeader("Accept-Encoding", "gzip, br")
```
Jorgesnchz: This class has a different type from the rest of additions `"Accept-Encoding"` instead of `HttpHeaders.ACCEPT_ENCODING`

Jorgesnchz (changes requested):
Code looks good, but I saw some things that I think could be changed:

I think that `br` decoder is registered only when `brotli` is on the classpath, and since it is not included in our package we might get an error, maybe we could add a test using a mock or a component test and see if the `br` encoding is really working or not

---

We could instead of 12 per-request `addHeader` calls, we could add it to the classes `ApiClient` and `SnapshotClient` to delete the extra code 


We could use a constant for `"gzip, br"`, since it is hardcoded in a lot of places

---

Maybe we could try to add some tests, either unit test or component tests

Documentation says "silently overwrites an `Accept-Encoding` default header with its own value", but the code does not overwrite, the compression exec runs first and only adds `if (!request.containsHeader(...))`, and `RequestDefaultHeaders` then skips a header name that is already present

We should also update the `docs/ref/modules/content-manager/architecture.md`

Jorgesnchz:
Also component test are broken: https://github.com/wazuh/wazuh-indexer-plugins/actions/runs/33635162978

plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/console/client/TelemetryClient.java
MauriFont: done

MauriFont:
It was decided to remove Brotli entirely.

AlexRuiz7 (approved):

PR #1523, opened 2026-09-03
Document how to change default Wazuh Indexer passwords

### LINKED ISSUE #1927: Install-time credential generation — indexer
## Description

The indexer's share of [#1928](https://github.com/wazuh/wazuh-indexer/issues/1928) — install-time credential generation. Read that epic first: the resolution ladder, `/etc/wazuh/credentials.env`, the install/start split and the upgrade and removal rules are defined there and are common to all three components. This issue is what the indexer has to build, plus the one thing only it has to do — upload the security configuration.

This issue began as the bug report that started the epic: a stock 5.0 AIO accepted `wazuh-demo:wazuh-demo`, `wazuh-admin:wazuh-admin`, `wazuh-readonly:wazuh-readonly`, `kibanaserver:kibanaserver`, `anomalyadmin:anomalyadmin` and `admin:admin` on `*:9200`, and `wazuh-demo` held Content Manager promote, space delete and Security Analytics admin. Part of that has since been fixed on `5.0.0` and the rest is what remains here.

**Already done on `5.0.0`, verified in the working tree:**

- `wazuh-demo` and `wazuh-admin` are gone from `distribution/src/config/security/internal_users.wazuh.yml`, which now ships only `wazuh-manager` and `wazuh-readonly`.
- `assemble.sh:242` removes the upstream demo accounts `anomalyadmin`, `kibanaro`, `logstash`, `readall` and `snapshotrestore` through `remove_security_entry`, which fails the build if an entry it expects has gone.
- `assemble.sh:279` removes the `own_index`, `kibana_user`, `readall`, `logstash` and `manage_snapshots` role mappings.

**What is still true, and is the work here:**

- `assemble.sh` deliberately keeps upstream's `admin` and `kibanaserver` entries, and both still carry the demo hash — password equals username in the shipped package.
- The two Wazuh accounts ship a fixed bcrypt hash each, under a header that reads `WARNING: Change these default passwords immediately after installation`. Every installation in the world gets the same two hashes.
- `install-demo-certificates.sh` issues a certificate for `CN=node-0.wazuh.indexer` with the SAN list `localhost`, `*.wazuh.indexer`, `RID.1 = 1.2.3.4.5`, `127.0.0.1` and `::1`, behind a `GENERATE_CERTS=true` gate, while `opensearch.prod.yml` hardcodes `CN=node-1,…` in `plugins.security.nodes_dn` and `CN=admin,…` in `authcz.admin_dn`.
- Nothing generates anything at install. There is no credential resolution at any point in the package's life.

`wazuh-readonly` is removed in this issue. The epic left it as an open decision; the decision is to drop it. It is the only discretionary account left, nothing in the product authenticates as it, and the goal is the smallest possible set of shipped accounts. After this issue the indexer ships exactly three: `admin`, `kibanaserver` and `wazuh-manager` — and none of them with a password.

### The resolver

- Ship `/usr/share/wazuh-indexer/bin/resolve-credentials`, one idempotent script implementing the epic's four-step ladder for the indexer, invoked unchanged from `postinst`, from `%post`, from the unit's `ExecStartPre` and from the container entrypoint in `wazuh/wazuh-docker`.
- Resolve the three accounts the indexer owns, each from its own key: `WAZUH_INDEXER_ADMIN_PASSWORD` → `admin`, `WAZUH_INDEXER_KIBANASERVER_PASSWORD` → `kibanaserver`, `WAZUH_INDEXER_MANAGER_PASSWORD` → `wazuh-manager`.
- Accept `DASHBOARD_PASSWORD` as an environment-only alias for `WAZUH_INDEXER_KIBANASERVER_PASSWORD`, for compatibility with existing `wazuh-docker` compose files.
- Publish all three keys into the managed block of `/etc/wazuh/credentials.env`, whether the value was generated or supplied, so the manager and the dashboard find them.
- Consume nothing: the indexer authenticates to no peer, so it can never reach step 3 and can always resolve on its own.
- Take step 0 from the marker `/var/lib/wazuh-indexer/.credentials-initialised`, created after the first successful resolution. It records that this installation created or accepted credentials and holds no secret.
- Bcrypt each plaintext locally with the security plugin's bundled `hash.sh`, reading the value from stdin and tolerating the JVM warning preamble it prints, then write the hash into `opensearch-security/internal_users.yml`.
- Generate a password of 32 characters over `A–Z a–z 0–9 . , _ + : @ % ^ = ~ -` with at least one lower-case letter, one upper-case letter and one digit by construction, satisfying PCI DSS v4.0 requirement 8.3.6; validate any supplied value against the same rule and stop at step 1 when it fails.

### Shipped accounts

- Remove `wazuh-readonly` entirely: its entry in `internal_users.wazuh.yml`, the `wazuh_readonly` role in `roles.wazuh.yml`, the `wazuh_readonly` mapping in `roles_mapping.wazuh.yml`, and the persona list in the `roles.wazuh.yml` header comment that names it.
- Remove every bcrypt hash from `internal_users.wazuh.yml`, and the `WARNING: Change these default passwords immediately after installation` header with them — after this issue no account ships with a password at all, so there is nothing to warn about.
- Extend `add_configuration_files()` in `assemble.sh` so the assembled `opensearch-security/internal_users.yml` carries no usable hash for `admin` or `kibanaserver` either. Both entries stay — they are needed and neither may be renamed — but the demo hashes they inherit from upstream's file must not survive the build.
- The package therefore ships exactly three internal users, `admin`, `kibanaserver` and `wazuh-manager`, none of which can authenticate until `resolve-credentials` has run.

### Security configuration upload

- Move the `securityadmin.sh` upload out of installation: it needs a running indexer, and under the epic the installer no longer starts one.
- Run it from an `ExecStartPost=+` step on `wazuh-indexer.service`, guarded by the marker file and by standalone detection, so it runs once on the first successful start of a standalone node.
- On a cluster node, print the command to the journal instead of running it: the upload must happen once, from one node, and the `.opendistro_security` index then wins over every node's local `internal_users.yml`.
- Consequently, indexer nodes 2..N need no password keys at all — state this in the requirement so it is not re-derived later.
- Rework `indexer-security-init.sh` accordingly, or have the new step call it. It currently resolves the admin certificate path from `http.pemtrustedcas` in `opensearch.yml` and aborts when `admin.pem` and `admin-key.pem` are not beside the root CA.

### Certificates

- Implement scenarios A to D from the epic for the indexer, driven entirely by what is present in the CA directory and in `/etc/wazuh-indexer/certs`, with no mode flag.
- Replace `install-demo-certificates.sh` and drop the `GENERATE_CERTS` gate in `postinst` and in the spec's `%post`: certificate resolution is not optional and is not a demo path.
- Derive the default SAN list from the hostname, the FQDN, loopback and the global addresses of default-route interfaces only, and let `WAZUH_INDEXER_CERT_SANS` replace it wholesale.
- Write the DN of the certificate it just minted into `plugins.security.nodes_dn`, and the admin DN into `authcz.admin_dn`, in the same step that mints it — `opensearch.prod.yml` hardcodes `CN=node-1,OU=Wazuh,O=Wazuh,L=California,C=US` and `CN=admin,…` today, and a node whose DN is not listed is rejected by the cluster.
- Log the SAN list and the DN used on every run, because a wrong certificate does not fail at start — the pre-start step opens no connections — and surfaces weeks later as a failed peer connection.

### Packaging

- `postinst` and `%post` run the full ladder, generate what the indexer owns, and still neither enable nor start the service; they exit `0` whatever they could not resolve, and print no secret.
- Add to `wazuh-indexer.service`: `ExecStartPre=+/usr/share/wazuh-indexer/bin/resolve-credentials`, and `StartLimitBurst=3` with `StartLimitIntervalSec=60` in `[Unit]`.
- Give the unit write access to `/etc/wazuh` and to the CA directory. `ProtectSystem=full` makes `/etc` read-only for the unit, and `/etc/wazuh-indexer` is writable only because of an explicit `ReadWritePaths=-/etc/wazuh-indexer`; the equivalent entries for the new paths must be added and the result verified, since the `+` prefix lifts `User=` but not the unit's mount namespace.
- On `purge`, remove the `WAZUH_INDEXER_*` keys from inside the managed block of `credentials.env` and the marker file, and leave the operator's own lines, the CA directory and `/etc/wazuh` alone.
- On a 5.x-to-5.x upgrade, change nothing: the marker makes step 0 true and any key still in the file is ignored.
- Add a `CHANGELOG.md` entry under `### Removed` for `wazuh-readonly` and under `### Added` for the resolver, each linking this issue.

- `admin` and `kibanaserver` may not be renamed or removed. `admin` is required by `securityadmin.sh` and cluster bootstrap; `kibanaserver` is hardcoded throughout the security plugin. `wazuh-manager` may not be removed either — the manager authenticates as it.
- Removing `wazuh-readonly` is a breaking change for any operator who mapped something to it. It must appear in the changelog under `### Removed` and be flagged for the release notes and for documentation.
- Use `remove_security_entry` for anything stripped in `assemble.sh` rather than a bespoke `sed`. It fails the build when an entry it expects is not there, which is the property that catches an upstream rename instead of shipping the grant.
- Do not take step 0 from `internal_users.yml`. It ships with the package, so inferring "already present" from it makes every upgrade and every restart a coin toss. The marker is the only valid test.
- Do not start or enable the service from `postinst` or `%post`, and do not abort either of them — aborting leaves the package half-configured, breaks `apt install -f` and fails image builds.
- Do not run `securityadmin.sh` from a maintainer script. The installer starts nothing, so there is no cluster to upload to.
- Do not print a secret to standard output. Maintainer script output lands in `/var/log/apt/term.log`, `/var/log/dnf.log`, the journal, Ansible and AWX output and CI logs. Write the file and print its path.
- Write every secret through stdin, never `argv` — `hash.sh`, the keystore tools and `securityadmin.sh` all accept it. A secret on a command line is world-readable in `ps`.
- Parse `credentials.env`, never source it; take `flock` for every write, because systemd starts units in parallel; validate the file and the CA directory for root ownership, mode and symlinks on every run, not only at creation.
- The `ExecStartPre` must carry the `+` prefix. Without it the step inherits `User=wazuh-indexer` and cannot read a `0600 root:root` file or write root-owned material — one character, and everything fails quietly.
- `StartLimitBurst` and `StartLimitIntervalSec` belong in `[Unit]`, not `[Service]`, on systemd 230 and later; without them a failing pre-start step under a restart policy retries forever and floods the journal.
- The pre-start step must open no network connection. Validation is presence and format only; depending on a peer being reachable breaks boot ordering and cluster restarts.
- Do not carry over the current SAN list. `*.wazuh.indexer` is a name every installation shares, so under any shared CA a node holding it can present a certificate for any other node; `RID.1 = 1.2.3.4.5` is a placeholder; and the loopback-only remainder would make a distributed install produce a certificate that is valid, installs cleanly and fails on the first peer connection.
- The CA private key must never leave the host that minted it, and must never be copied into a certificates bundle. The bootstrap CA is local and disposable: a host that minted its own and later joins a real cluster does not merge trust, the cluster's CA reissues everything.
- Every password, generated or supplied, satisfies PCI DSS v4.0 requirement 8.3.6 — 12 to 64 characters with both letters and digits. An invalid supplied value is a failure, not an absence: print which key was rejected and against which rule, never the value.
- Lands on `5.0.0`. The container entrypoint wiring is in `wazuh/wazuh-docker` and is out of scope here beyond shipping a script it can call.

## Plan

- [ ] Remove `wazuh-readonly` from `internal_users.wazuh.yml`, `roles.wazuh.yml`, `roles_mapping.wazuh.yml` and the header comment
- [ ] Strip every bcrypt hash and the change-your-password warning from `internal_users.wazuh.yml`
- [ ] Extend `add_configuration_files()` in `assemble.sh` so `admin` and `kibanaserver` ship with no usable hash
- [ ] Write `resolve-credentials`: ladder, `credentials.env` parsing and locking, path validation, generation, `hash.sh` over stdin, marker file
- [ ] Wire it into `postinst` and the spec's `%post`; drop the `GENERATE_CERTS` gate
- [ ] Add `ExecStartPre=+`, the start limit and the `ReadWritePaths` entries to `wazuh-indexer.service`; verify the pre-start step can write `/etc/wazuh` under `ProtectSystem=full`
- [ ] Move the `securityadmin.sh` upload to `ExecStartPost=+` with marker and standalone guards; print the command on a cluster node; rework `indexer-security-init.sh`
- [ ] Implement certificate scenarios A to D, the derived SAN list and `WAZUH_INDEXER_CERT_SANS`; retire `install-demo-certificates.sh`
- [ ] Write the minted DN into `plugins.security.nodes_dn` and the admin DN into `authcz.admin_dn`; remove the hardcoded values from `opensearch.prod.yml`
- [ ] Handle `purge`: own keys inside the managed block, plus the marker
- [ ] Verify a 5.x-to-5.x upgrade changes nothing
- [ ] Verify on DEB and RPM: fresh AIO, distributed node, node 2..N with no password keys, and a node started with a missing credential
- [ ] `CHANGELOG.md` entries under `### Removed` and `### Added`
- [ ] Flag the `wazuh-readonly` removal for release notes and documentation

AlexRuiz7:
## Description

Document how to change default Wazuh Indexer passwords

Closes wazuh/wazuh-indexer#1927 

## Proposed Changes

Documentation.

### Results and Evidence

None

### Artifacts Affected

Documentation

### Configuration Changes

None

### Documentation Updates
 
Yes

### Tests Introduced

None

## Review Checklist

- [ ] Code changes reviewed
- [ ] ...


docs/ref/getting-started/installation.md
```diff
@@ -253,3 +255,123 @@ Run the Wazuh indexer `indexer-security-init.sh` script on any Wazuh indexer nod
...
+03/09/2026 12:25:15 INFO: The password for user wazuh-admin is Qk7Rm2wZpL9xTvNb.4sHdY*eA6cJgW3u
+```
+
+Each run:
+
+- Writes a timestamped copy of the previous user definitions to `/etc/wazuh-indexer/internalusers-backup/`.
+- Applies the new password to the running cluster and persists it to `/etc/wazuh-indexer/opensearch-security/internal_users.yml`, so the change survives a later run of `indexer-security-init.sh`.
+- Logs to `/var/log/wazuh-passwords-tool.log`.
+
+The tool authenticates with the admin certificate at `/etc/wazuh-indexer/certs/admin.pem` and its key at `/etc/wazuh-indexer/certs/admin-key.pem`. If you deployed the certificates elsewhere, pass `-c` and `-k` with the correct paths.
```
felipegonzalezmv: -c and -k were removed in [a3d1a97](https://github.com/wazuh/wazuh-installation-assistant/commit/a3d1a972) ("remove admin key custom paths options"). On current 5.0.0 the tool only accepts -a -A -au -ap -u -p -v -h, so this command prints the help and exits 1 without changing the password:

felipegonzalezmv (changes requested):

docs/ref/getting-started/installation.md
AlexRuiz7: Good catch. Solved.

felipegonzalezmv (approved):
LGTM 🟢

PR #1545, opened 2026-09-07
Adapt logtest to percolator-based rule evaluation

### LINKED ISSUE #321: [BUG] `logtest` reports `contains` matches that never fire in the real ingestion pipeline (case sensitivity)
## Summary

The `contains` modifier is inconsistent between the standard ingestion pipeline and `logtest`. Real ingestion is case-sensitive, exactly as documented; `logtest`'s detection stage is not, and reports a match that the deployed detector never actually produces. Anyone validating a rule through `logtest` alone will see it "work" when it doesn't.

## Reproduction

Two events, differing only in the case of `SELECT`, both fed through the standard ingestion pipeline:

```
203.0.113.11 - - [04/Sep/2026:10:16:00 +0000] "GET /item.php?id=-1 UNION SELECT username,password FROM users-- HTTP/1.1" 200 340 "-" "sqlmap/1.7"

203.0.113.11 - - [04/Sep/2026:10:16:00 +0000] "GET /item.php?id=-1 UNION sElect username,password FROM users-- HTTP/1.1" 200 340 "-" "sqlmap/1.7"
```

The first line correctly generates a finding — it matches the `UNION SELECT` branch of the rule's `contains` list. The second does not generate any finding in production. But `logtest` reports the second line as detected too:

Image

Rule condition under test:

```yaml
detection:
  selection:
    http.request.method: "GET"
  selection_1:
    url.original|contains:
      - "@@version"
      - "%271%27%3D%271"
      - "=select "
      - "=select("
      - "=select%20"
      - "=select+"
      - "concat_ws("
      - "CONCAT(0x"
      - "from mysql.innodb_table_stats"
      - "from%20mysql.innodb_table_stats"
      - "from+mysql.innodb_table_stats"
      - "group_concat("
      - "information_schema.tables"
      - "json_arrayagg("
      - "or 1=1#"
      - "or%201=1#"
      - "or+1=1#"
      - "order by "
      - "order%20by%20"
      - "order+by+"
      - "select * "
      - "select database()"
      - "select version()"
      - "select%20*%20"
      - "select%20database()"
      - "select%20version()"
      - "select+*+"
      - "select+database()"
      - "select+version()"
      - "select%28sleep%2810%29"
      - "SELECTCHAR("
      - "table_schema"
      - "UNION ALL SELECT"
      - "UNION SELECT"
      - "UNION%20ALL%20SELECT"
      - "UNION%20SELECT"
      - "UNION+ALL+SELECT"
      - "UNION+SELECT"
      - "'1'='1"
  filter_main_status:
    http.response.status_code: 404
  condition: selection and selection_1 and not 1 of filter_main_*
```

## Possible root cause on the ingestion side (confirmed against source)

`contains` compiles to `Operator::CN` in `opBuilderHelperFilter.cpp`, whose comparison function is a raw substring search with no case folding:

```cpp
case Operator::CN:
    cmpFunction = [](std::string_view l, std::string_view r)
    {
        if (!r.empty())
        {
            return l.find(r) != std::string_view::npos;
        }
        return false;
    };
    break;
```

## Ask

We need consistent application of the modifiers between the logtest and the standard ingestion pipeline.

It's preferrable that the `contains` modifier is case-insensitive. This will enable coverage for basic obfuscation techniques that rely on case manipulation.

Jorgesnchz:
## Description

This PR is a companion of this PR: https://github.com/wazuh/wazuh-indexer-security-analytics/pull/324

This PR is the Content Manager half: it hands Security Analytics what it needs to evaluate rules the way a deployed detector does, reports the new "could not evaluate" outcome instead of hiding it, and, so the behavior is actually covered by
tests, stops mocking rule evaluation in the integration-test cluster.

Closes https://github.com/wazuh/wazuh-indexer-security-analytics/issues/321

## Proposed Changes

**1. Detector coordinates are passed to the evaluator.** Security Analytics now compiles rules and percolates them, so it needs to know which percolator index to use and which source indices the compiled queries must resolve their fields against. `SecurityAnalyticsService#evaluateRulesAsync` takes three more arguments, `integrationId`, `logType`, `sourceIndices`, and `WEvaluateRulesRequest` carries them.

`LogtestService#extractDetectorTarget` reads them off the integration document logtest has already fetched, deliberately mirroring how a detector is built from that same document

**2. Failures say what happened.** 
**3. `plugins.content_manager.security_analytics.mock` (new).** One flag, `plugins.content_manager.engine.mock`, used to swap out *both* the Engine and the Security Analytics service. 

### Results and Evidence

Full runs in `pr-321-evidence.md`.

- **Unit tests**: :white_check_mark: 
- **Integration tests**:  :white_check_mark: 
- **Live validation on the AIO** In: https://github.com/wazuh/wazuh-indexer-security-analytics/pull/324#issue-5374292360

### Artifacts Affected

Content Manager plugin

### Configuration Changes

- **New**: `plugins.content_manager.security_analytics.mock` (node scope, filtered). Defaults to the
  value of `plugins.content_manager.engine.mock`, so existing deployments and dev setups are
  unaffected.

### Documentation Updates

- `docs/ref/modules/ruleset-management/rules.md`
- `docs/ref/modules/content-manager/rule-testing.md` 
- `docs/ref/modules/content-manager/api.md` and `plugins/content-manager/openapi.yml`
- `docs/dev/plugins/content-manager-logtest.md` 
- `openapi.yml` 

### Tests Introduced

| Test | Covers |
| --- | --- |
| `LogtestIT#testDetectionPercolatesRulesLikeADetector` | The reported case end to end: the exact-case event matches, the mixed-case event does **not** — which is what a deployed detector does — an unrelated event does not match, and a rule over a field the source index does not map is reported under `skipped` with a reason rather than silently counted as a non-match |

`LogtestServiceTests` was updated for the wider `evaluateRulesAsync` signature, and its integration
fixtures now carry the `category` and `metadata.title` a real integration document has — without them
the service correctly refuses to evaluate.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,6 +188,7 @@
 - [BUG] Retry sending resources to SAP after failure [(#1487)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1487)
 - [BUG] `privacy_default_per_provider` mapped as object under `dynamic:strict` rejects new provider overrides [(#1498)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1498)
 - [BUG] Resource creation requires `indices:admin/create` on `.wazuh-content-manager-resource-locks` for the calling user [(#1520)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1520)
+- [BUG] `logtest` reports `contains` matches that never fire in the real ingestion pipeline (case sensitivity) [(#321)](https://github.com/wazuh/wazuh-indexer-security-analytics/issues/321)
 
 ## Prior versions
 - []()
--- a/docs/dev/plugins/content-manager-logtest.md
+++ b/docs/dev/plugins/content-manager-logtest.md
@@ -66,7 +66,7 @@ Responsibilities:
 2. Validates the required fields `space`, `integration`, and `input`.
 3. Validates that `space` is not `"draft"`.
 4. Validates that `input` is a JSON object (not a string or array).
-5. Delegates to `LogtestService.executeDetection(integrationId, space, inputEvent)`.
+5. Delegates to `LogtestService.executeDetectionAsync(integrationId, space, inputEvent)`.
 
 All three transport actions run this work on the dedicated `content_manager_logtest` thread pool rather than the transport thread (see [Request size limit and concurrency](#request-size-limit-and-concurrency)).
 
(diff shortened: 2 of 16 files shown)


ferooo23 (dismissed):
LGTM!
Validation: https://github.com/wazuh/wazuh-indexer-security-analytics/pull/324#issuecomment-5587640546

AlexRuiz7 (changes requested):
Revert logtest's response to the previous version:

- no skipped
- no AND

Those changes are out of the scope of the issue and violate the contract adopted in https://github.com/wazuh/wazuh-indexer-security-analytics/issues/56

ferooo23 (dismissed):

AlexRuiz7 (approved):

PR #1547, opened 2026-09-07
Add permissions check on endpoint to set subscription to CTI

felipegonzalezmv:
## Description

CTI registration is an OAuth device flow, and authorization is only evaluated at the very
end of it: the user approves in the CTI console — which already **creates the environment
on the CTI side** — and only then does the indexer get to answer `403`. That leaves an
orphaned environment, and gives the dashboard no way to know in advance whether the
current user may register.

This PR lets a caller ask **"am I allowed to register a CTI subscription?"** *before*
starting the flow, via `perform_permission_check=true` on
`POST /_plugins/_content_manager/subscription`.

Closes #6224

## Proposed Changes

**We do not implement the check.** `perform_permission_check` is a native OpenSearch
parameter: the security plugin evaluates the action and answers with its own
`PermissionCheckResponse` *instead of executing it*, so there are no side effects. Our
only job was to stop getting in the way. Three things blocked it:

| Obstacle | Fix |
|---|---|
| The handler always parsed the body ⇒ `400 request body is required` | No body is read in check mode |
| `validate()` rejected the missing token, and validation runs **before** the action filters ⇒ `400` before the filter ever saw it | `validate()` skips the token check in check mode |
| The listener was typed on `MessageStatusResponse` ⇒ `ClassCastException` ⇒ **`500`** | Listener typed on `ActionResponse`, discriminating with `instanceof StatusToXContentObject` |

### The contract

```http
POST /_plugins/_content_manager/subscription?perform_permission_check=true
```

No body, no access token, no side effects, idempotent. **Always answers `200`** — permitted
and denied differ only in the body:

```json
{ "accessAllowed": true,  "missingPrivileges": [] }
{ "accessAllowed": false, "missingPrivileges": ["cluster:admin/content_manager/subscription/create"] }
```

**Clients must branch on `accessAllowed`, not on the HTTP status.**

Without the parameter, behaviour is unchanged.

### Results and Evidence

Before this PR the endpoint could not answer at all: `400` without a body, and **`500
class_cast_exception`** with one. That `500` is what proved the security filter *was*
working and our listener was the thing breaking.

**1. Permitted** — as `wazuh-admin`:

```
POST _plugins/_content_manager/subscription?perform_permission_check=true
```
Expected `200` → `{"accessAllowed": true, "missingPrivileges": []}`

<details><summary>Details</summary>
<p>

image

</p>
</details> 

**2. Denied** — as `wazuh-readonly`, same request.

Expected **`200` (not `403`)** → `{"accessAllowed": false, "missingPrivileges": ["cluster:admin/content_manager/subscription/create"]}`

<details><summary>Details</summary>
<p>

image

</p>
</details> 

**3. No side effects** — `GET _plugins/_content_manager/subscription` before and after.

Expected `is_registered` stays `false`.

<details><summary>Details</summary>
<p>

image

</p>
</details> 

**4. Backwards compatible** — as `wazuh-readonly`, without the parameter:

```
POST _plugins/_content_manager/subscription
{ "access_token": "dummy" }
```
Expected `403 security_exception`, same as before this PR.

<details><summary>Details</summary>
<p>

image

</p>
</details> 

Also verified: the security-disabled fallback (`plugins.security.disabled=true`) returns
the same `accessAllowed: true` body, and the full 16-probe before/after set is green.

### Artifacts Affected

`wazuh-indexer-content-manager` plugin only. No packaging, no config files, no changes to
`roles.wazuh.yml` or `action_groups.wazuh.yml`.

### Configuration Changes

None. One new optional query parameter (`perform_permission_check`, boolean, default
`false`), owned by the OpenSearch security plugin. Fully backwards compatible.

### Documentation Updates

- `plugins/content-manager/openapi.yml` — parameter, `200` response and
  `PermissionCheckResponse` schema, with permitted/denied examples.
- Javadoc on `RestPostSubscriptionAction`.

### Tests Introduced

**19 new unit tests.** Full suite: **745 tests, 0 failures.**

| Class | New | Scope |
|---|---|---|
| `IndexSubscriptionRequestTests` *(new)* | 9 | Both request shapes, validation skip, wire round trip |
| `RestPostSubscriptionActionTests` | +9 | Parameter parsing and all three response paths |
| `TransportIndexSubscriptionActionTests` | +1 | Security-disabled fallback never calls `register` |

`PermissionCheckResponse` can't be referenced from a test — it lives in the security
plugin's classloader, which is exactly what the `class_cast_exception` reports — so the
tests use a double with the same two relevant properties.

### Known limitation

`GET` and `DELETE /subscription` still return `500` with this parameter, for the same
listener-typing reason. **Pre-existing, not a regression.** Left out because it needs its
own decision: what a permission check means on `GET` (where `wazuh_readonly` *does* hold
the permission), and whether a shared listener helper beats patching routes one by one.
Worth a follow-up issue.

## Review Checklist

- [ ] Code changes reviewed
- [ ] CHANGELOG entry added, or `no-changelog` label applied
- [ ] Follow-up issue opened for the `GET`/`DELETE` limitation

### DIFF
--- a/docs/ref/modules/content-manager/api.md
+++ b/docs/ref/modules/content-manager/api.md
@@ -96,6 +96,8 @@ YAML parsing preserves numeric type fidelity. Floating-point values like `5.0` a
 
 Stores the provided CTI access token in the `.wazuh-internal-state` hidden index and loads it into memory. If the index does not exist it is recreated automatically before writing.
 
+> To find out whether the current user is allowed to register, without registering, see [Check registration permission](#check-registration-permission).
+
 #### Request
 
 - Method: `POST`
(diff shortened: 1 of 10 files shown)


MauriFont (dismissed):
## Manual verification

I manually verified the permission-check behavior against a running instance:

**Permitted — `admin` role:**
curl -sk -u admin:admin -X POST "https://localhost:9200/_plugins/_content_manager/subscription?perform_permission_check=true"
→ {"accessAllowed":true,"missingPrivileges":[]}

Image

**Denied — `wazuh-readonly` role:**
curl -sk -u wazuh-readonly:wazuh-readonly -X POST "https://localhost:9200/_plugins/_content_manager/subscription?perform_permission_check=true"
→ {"accessAllowed":false,"missingPrivileges":["cluster:admin/content_manager/subscription/create"]}

Image

Both cases return `200`, and the response body correctly reflects the caller's actual privileges. LGTM.

AlexRuiz7 (changes requested):
Missing documentation for [MdBook](https://wazuh.github.io/wazuh-indexer-plugins/ref/modules/content-manager/api.html#subscription-management).

felipegonzalezmv:
> Missing documentation for [MdBook](https://wazuh.github.io/wazuh-indexer-plugins/ref/modules/content-manager/api.html#subscription-management).

Resolved here: https://github.com/wazuh/wazuh-indexer-plugins/pull/1547/changes/1747fe552da79f7ecc7d1545c8411030f945f5d2

AlexRuiz7 (approved):

PR #1554, opened 2026-09-08
Mediate AI assistant session writes through the setup plugin

felipegonzalezmv:
## Description

`wazuh_ai_assistant` is mapped to `users: ["*"]` and grants index-level `write` on
`wazuh-ai-assistant-sessions*` with no scoping. Document Level Security is a read-path filter and
cannot scope a write, so the `dls` key on the sibling `read` block protects nothing on the write
path: any account reaching the index can store a session whose `user` field names somebody else.
The document then surfaces in the victim's history and is invisible to its author. The shipped
**read-only** profile is enough to do it.

No role syntax can express "this document's `user` must equal your name", so `user` has to stop
being client input. Session writes are now mediated by the setup plugin, which derives the owner
from the authenticated principal and discards whatever the body said.

Part of internal-devel-requests#6111.

## Proposed Changes

Four write routes, one new cluster permission (`plugin:wazuh/ai_assistant/session/write`). **No read
routes** — listing and reading stay direct queries under the role's existing DLS filter, which
scopes reads correctly and already gives search, sorting and pagination.

| Operation | Route |
| --- | --- |
| `createAiAssistantSession` | `POST /_plugins/_setup/ai_assistant/sessions` |
| `updateAiAssistantSession` | `PUT /_plugins/_setup/ai_assistant/sessions/{id}` |
| `renameAiAssistantSession` | `PATCH /_plugins/_setup/ai_assistant/sessions/{id}` |
| `deleteAiAssistantSession` | `DELETE /_plugins/_setup/ai_assistant/sessions/{id}` |

`TransportPutAiAssistantSessionAction` resolves the caller from the security plugin's thread-context
transient *before* stashing the context, stamps it onto `user`, then writes with the plugin's own
privileges. Update, rename and delete resolve the stored document through a lookup filtered on that
same owner, so somebody else's session is simply not found.

**This PR is inert on a stock cluster and does not close the issue on its own:** no shipped role
grants the permission yet, so the endpoints and the old direct-write path coexist and nothing
breaks.

<details><summary><b>Security decisions worth a reviewer's attention</b></summary>

- **`404`, not `403`, for a session owned by someone else.** A `403` would confirm the id exists.
  The ownership filter is part of the lookup query, so the `404` falls out of it.
- **No impersonation and no administrative override**, for `all_access` or `wazuh-admin` — this
  preserves what the DLS query already did, which applies to `admin` too.
- **Server-owned fields are dropped, not rejected.** `user`, `created_at`, `updated_at`,
  `@timestamp` and unknown keys are discarded silently, so a client can `POST` back a document it
  previously read without a `400`.
- **The caller is read from the transient directly** rather than through
  `org.opensearch.commons.authuser.User`, to avoid pulling `common-utils` into a plugin that
  deliberately carries almost no dependencies. Worth revisiting if a reviewer prefers the dependency.
- **With security disabled the owner resolves to the `_shared` sentinel** — the same one the
  Dashboard uses — treated as a real owner so the path is exercised rather than bypassed.

</details>

<details><summary><b>Three data-stream constraints that shaped the implementation</b></summary>

Platform behaviour, not choices. All three look like defects to a later reader, so they are
commented where they bite.

| Constraint | Consequence |
| --- | --- |
| A get-by-id cannot be used — the backing index rolls over daily and the get API targets one concrete index | `findHit` searches the stream, which reports the backing index plus the `seq_no`/`primary_term` the next write needs, in one round trip |
| The partial `_update` API is refused wherever DLS applies (`security_exception: Update is not supported when FLS or DLS or Fieldmasking is activated`) | Update and rename are a full replace against the backing index, never the stream name, which accepts appends only |
| A backing index rejects an unconditional `op_type=index` write (`illegal_argument_exception: ... no if_primary_term and if_seq_no set targeting backing indices is disallowed`) | Optimistic concurrency is mandatory — every write carries a pair, surfaced to clients as the opaque `version` token |

A version conflict is `409` and **never retried**: retrying with a freshly read pair is precisely
the silent overwrite `expected_version` exists to prevent.

Two smaller rules, both deliberate and both easy to "fix" back into a bug:

- **`PATCH` does not re-stamp `updated_at`.** A rename is not session activity; bumping it would
  jump the row to the top of a list ordered by last activity.
- **`PUT`'s `title` is optional.** A chat client auto-saves every turn, and resending a recomputed
  title would silently revert a rename the user had just made.

</details>

<details><summary><b>Scope and rollout order</b></summary>

Not in this PR:

- **No change to `roles.wazuh.yml`, `action_groups.wazuh.yml` or `roles_mapping.wazuh.yml`** —
  those live in `wazuh-indexer`. Adding the grant and removing the `write` block is what actually
  closes the issue.
- **No Dashboard change.** Its three write functions in
  `plugins/wazuh-ai-assistant/server/conversation-store.ts` still write directly; its read
  functions stay as they are.

| # | Change | Repo | Breaks anything? |
| --- | --- | --- | --- |
| 1 | Endpoints + docs + tests (**this PR**) | `wazuh-indexer-plugins` | No — both paths work |
| 2 | Dashboard moves its three writes onto the endpoints | `wazuh-dashboard` | No |
| 3 | Delete the `write` block, add the cluster permission | `wazuh-indexer` | Breaks any client still writing directly |

Step 3 must be gated on step 2, or sessions stop persisting silently.

</details>

### Results and Evidence

Live security-enabled cluster: OpenSearch 3.6.0, `wazuh-indexer-setup 5.0.0.0`, stock Wazuh security
config with step 3 applied to the test cluster only. Both identities are **stock users**, so this is
reproducible with no setup — and both hold `wazuh_ai_assistant` only because it is mapped to `*`:

| User | Roles |
| --- | --- |
| `wazuh-readonly` | `wazuh_readonly`, `own_index`, **`wazuh_ai_assistant`** |
| `admin` | `all_access`, `wazuh_admin`, `own_index`, **`wazuh_ai_assistant`** |

`admin` is a genuine second identity here, not a superuser shortcut: the per-owner DLS applies to it
too.

**The headline result.** `wazuh-readonly` creates a session forging `"user": "admin"`, and what
lands in the index is:

```json
{ "user": "wazuh-readonly", "title": "forged ownership attempt", ... }
```

Not `admin`. That one line is the fix.

| Check | Before | After |
| --- | --- | --- |
| `POST /wazuh-ai-assistant-sessions/_doc` forging `user: admin`, as `wazuh-readonly` | **`201 created`** | **`403`** `no permissions for [indices:data/write/index]` |
| Victim's `_search {"term":{"user":"admin"}}` | forged session listed as their own | not present (`total: 0`) |
| Author's `_count` of what they just wrote | `0` — DLS hid it from them | n/a, the write is refused |
| `POST …/ai_assistant/sessions` forging `user: admin` | route absent | `200`, stored as `wazuh-readonly` |
| `PUT`/`PATCH`/`DELETE` on another user's session | n/a | `404`, session untouched |
| Endpoint without the cluster permission | n/a | `403` `cluster:admin/ai_assistant/session/write` |

Requests below are shown as Dev Tools console blocks
(<https://localhost:5601/app/dev_tools#/console>); the console proxies `PATCH`, so all four routes
are reproducible there. **Dev Tools sends requests as the logged-in user**, so the cross-user steps
need a re-login or curl — otherwise you run both sides as the same user and prove nothing.

<details><summary><b>1 — Full conversation lifecycle: create → add turns → rename → delete</b></summary>

All as `wazuh-readonly`.

**1.1 Create with the first user turn.**

```
POST /_plugins/_setup/ai_assistant/sessions
{
  "title": "Any brute force attempts today?",
  "messages": [
    { "role": "user", "content": "Any brute force attempts today?", "createdAt": 1788961419723 }
  ]
}
```

```json
// 200
{
  "id": "OaUIh6ABB7kcGdordSSl",
  "title": "Any brute force attempts today?",
  "created_at": "2026-09-09T16:37:53.701003433Z",
  "updated_at": "2026-09-09T16:37:53.701003433Z",
  "messages": [ { "role": "user", "content": "Any brute force attempts today?", "createdAt": 1788961419723 } ],
  "version": "47:2"
}
```

`200`, not `201`, matching `POST /ai_assistant/providers`. `created_at == updated_at` on create, and
`version` is returned so the first `PUT` needs no intervening read.

**1.2 Add the assistant's reply — the auto-save after turn 1, sending back 1.1's version.**

```
PUT /_plugins/_setup/ai_assistant/sessions/OaUIh6ABB7kcGdordSSl
{
  "messages": [
    { "role": "user", "content": "Any brute force attempts today?", "createdAt": 1788961419723 },
    { "role": "assistant", "content": "No brute force activity in the last 24 hours.",
      "createdAt": 1788961419744, "providerName": "anthropic",
      "providerModel": "claude-opus-5", "privacyEnabled": false }
  ],
  "expected_version": "47:2"
}
```

```json
// 200
{ "created_at": "2026-09-09T16:37:53.701003433Z",
  "updated_at": "2026-09-09T16:37:54.753909910Z",
  "title": "Any brute force attempts today?",
  "messages": [ /* 2 entries */ ], "version": "48:2" }
```

`created_at` carried over, `updated_at` advanced, `title` kept (none sent), fresh `version`. The
provider-specific message fields round-trip untouched: `messages` is mapped
`{"type": "object", "enabled": false}`, so a provider returning unanticipated fields cannot break
saving.

**1.3 Second turn, using the version the last write returned** — the loop a chat client actually runs.

```
PUT /_plugins/_setup/ai_assistant/sessions/OaUIh6ABB7kcGdordSSl
{ "messages": [ /* 4 entries: two full turns */ ], "expected_version": "48:2" }
```

```json
// 200
{ "updated_at": "2026-09-09T16:37:55.738418875Z", "version": "50:2", ... }
```

**1.4 Rename.**

```
PATCH /_plugins/_setup/ai_assistant/sessions/OaUIh6ABB7kcGdordSSl
{ "title": "  Agent 003 connectivity  " }
```

```json
// 200
{ "id": "OaUIh6ABB7kcGdordSSl", "title": "Agent 003 connectivity",
  "updated_at": "2026-09-09T16:37:55.738418875Z", "version": "50:2" }
```

Title **trimmed**; `updated_at` **byte-identical to 1.3's**, so the session does not jump the list.
The fresh `version` means a client renaming the session it has open does not `409` on its next
auto-save.

**1.5 Read it back with a direct query** — the way the Dashboard reads, since there is no read
endpoint.

```
GET /wazuh-ai-assistant-sessions/_search
{ "query": { "ids": { "values": ["OaUIh6ABB7kcGdordSSl"] } }, "seq_no_primary_term": true }
```

```json
// 200
{ "hits": { "total": { "value": 1 }, "hits": [ {
  "_index": ".ds-wazuh-ai-assistant-sessions-000003",
  "_seq_no": 50, "_primary_term": 2,
  "_source": {
    "@timestamp": "2026-09-09T16:37:53.701003433Z",
    "created_at": "2026-09-09T16:37:53.701003433Z",
    "updated_at": "2026-09-09T16:37:55.738418875Z",
    "title": "Agent 003 connectivity",
    "user": "wazuh-readonly",
    "messages": [ /* 4 entries, all provider fields intact */ ]
  } } ] } }
```

`user` is the caller. `@timestamp` equals `created_at` — it exists only for the data stream's
bookkeeping and is never exposed by the API. `_seq_no: 50` / `_primary_term: 2` is where the
`"50:2"` token comes from, which is how a client recovers from a `409`.

**1.6 Delete.**

```
DELETE /_plugins/_setup/ai_assistant/sessions/OaUIh6ABB7kcGdordSSl
```

```json
// 200
{ "message": "Session deleted.", "status": 200, "id": "OaUIh6ABB7kcGdordSSl" }
```

`DELETE` is the one route that keeps the house `{message, status, id}` envelope, having nothing to
round-trip.

**1.7 It is gone, and a repeated delete is `404`.**

```json
// GET _search -> 200
{ "hits": { "total": { "value": 0 }, "hits": [] } }

// DELETE again -> 404
{ "error": { "type": "resource_not_found_exception",
             "reason": "Session not found: OaUIh6ABB7kcGdordSSl" }, "status": 404 }
```

</details>

<details><summary><b>2 — The issue's actual claim: another user's <code>user</code> can no longer be set</b></summary>

**2.1 Create a session forging `"user": "admin"`** — the request the issue is about, as
`wazuh-readonly`.

```
POST /_plugins/_setup/ai_assistant/sessions
{
  "title": "forged ownership attempt",
  "messages": [],
  "user": "admin",
  "created_at": "2000-01-01T00:00:00Z",
  "junk": "dropped"
}
```

```json
// 200
{ "id": "OqUIh6ABB7kcGdoriSS3", "title": "forged ownership attempt",
  "created_at": "2026-09-09T16:37:58.839386747Z",
  "updated_at": "2026-09-09T16:37:58.839386747Z",
  "messages": [], "version": "52:2" }
```

No `400` — the forged keys are dropped, not rejected, so a round-tripped document still saves. And
`created_at` is *now*, not the `2000-01-01` sent.

**2.2 What was actually stored.** Read as the owner, since `admin` is subject to the same DLS.

```json
// 200
{ "hits": { "total": { "value": 1 }, "hits": [ { "_source": {
  "user": "wazuh-readonly",
  "title": "forged ownership attempt",
  "created_at": "2026-09-09T16:37:58.839386747Z",
  "updated_at": "2026-09-09T16:37:58.839386747Z"
} } ] } }
```

**`"user": "wazuh-readonly"`, not `"admin"`.** The `junk` key never reached the index either.

**2.3 `admin`'s own history does not contain it.**

```
GET /wazuh-ai-assistant-sessions/_search
{ "query": { "term": { "user": "admin" } }, "_source": ["user", "title"] }
```

```json
// 200
{ "hits": { "total": { "value": 0 }, "hits": [] } }
```

Before the fix, this same pair of steps returned the forged session here, listed as `admin`'s own.

**2.4–2.6 `admin` creates a conversation** (`O6UIh6ABB7kcGdorjSTb`), then `wazuh-readonly` tries to
overwrite, rename and delete it.

```
PUT    /_plugins/_setup/ai_assistant/sessions/O6UIh6ABB7kcGdorjSTb
{ "messages": [ { "role": "user", "content": "injected by wazuh-readonly" } ] }

PATCH  /_plugins/_setup/ai_assistant/sessions/O6UIh6ABB7kcGdorjSTb
{ "title": "hijacked" }

DELETE /_plugins/_setup/ai_assistant/sessions/O6UIh6ABB7kcGdorjSTb
```

```json
// 404, all three
{ "error": { "type": "resource_not_found_exception",
             "reason": "Session not found: O6UIh6ABB7kcGdorjSTb" }, "status": 404 }
```

**2.7 `admin`'s conversation is untouched** — checked explicitly, because a status code alone would
not catch a refusal that still wrote.

```json
// 200
{ "hits": { "total": { "value": 1 }, "hits": [ { "_source": {
  "user": "admin", "title": "admin private conversation",
  "messages": [ { "role": "user", "content": "internal", "createdAt": 1788961419723 } ]
} } ] } }
```

**2.8 No administrative override either.** `admin` holds `all_access` and still cannot reach the
other user's session:

```
PUT    /_plugins/_setup/ai_assistant/sessions/OqUIh6ABB7kcGdoriSS3
DELETE /_plugins/_setup/ai_assistant/sessions/OqUIh6ABB7kcGdoriSS3
```

```json
// 404, both
{ "error": { "type": "resource_not_found_exception",
             "reason": "Session not found: OqUIh6ABB7kcGdoriSS3" }, "status": 404 }
```

**2.9 The original proof-of-concept from the issue, verbatim.**

```
POST /wazuh-ai-assistant-sessions/_doc
{ "user": "admin", "title": "INJECTED BY wazuh-readonly",
  "@timestamp": "2026-09-09T09:00:00Z",
  "created_at": "2026-09-09T09:00:00Z", "updated_at": "2026-09-09T09:00:00Z" }
```

```json
// 403
{ "error": { "type": "security_exception",
    "reason": "no permissions for [indices:data/write/index] and User [name=wazuh-readonly, backend_roles=[], requestedTenant=null]" },
  "status": 403 }
```

Before the change this returned `201 created`, landing in
`.ds-wazuh-ai-assistant-sessions-000003` under `"user": "admin"`.

</details>

<details><summary><b>Regression probes, and two consequences worth knowing</b></summary>

The old path must keep working until step 3 lands:

| Probe | Result |
| --- | --- |
| `_search` on the stream as `wazuh-readonly` after removing the `write` block | **`200`** — a `_search` needs index-level `read`, which the surviving block grants, and no cluster permission |
| Per-owner DLS still isolates reads | `wazuh-readonly` sees only its own; `admin` sees only its own |
| Direct `_doc` write with the `write` grant still in place | `201` — nothing breaks for a client that has not migrated |

Two consequences that only show up in teardown, worth knowing before someone files them as bugs:

- **After step 3, an ordinary user cannot `_delete_by_query` its own sessions** — no index-level
  write remains, so the endpoint is the only way to remove one. `admin` cannot do it on their behalf
  either, because the DLS hides those documents from it.
- **A "no leftovers" check has to be run per identity.** An `admin`-only count is DLS-scoped to
  admin's own documents and reports a clean cluster that is not one.

</details>

### Artifacts Affected

`wazuh-indexer-setup` plugin only — eleven new classes plus `SetupPlugin` wiring. No packaging
change, no new dependency, no new plugin setting, no index-template or mapping change, and no
security-config change in this repository.

### Configuration Changes

One new cluster permission, `cluster:admin/ai_assistant/session/write`, exposed as the action group
`plugin:wazuh/ai_assistant/session/write` when the `wazuh-indexer` side lands.

**Additive and inert.** New routes on a path prefix that had none, no existing route changed, no
default altered. Until a role grants the permission, every call answers `403` and nothing else
behaves differently.

<details><summary>Behaviour when the new optional parameters are absent</summary>

- `expected_version` absent narrows the concurrency check to "since this request started" rather
  than disabling it — the platform requires a version pair either way. An undecodable token is
  treated as absent rather than rejected.
- `title` absent on `PUT` means "keep the stored title".

</details>

### Documentation Updates

<details><summary>Five files, both layers</summary>

| File | Layer |
| --- | --- |
| `plugins/setup/openapi.yml` | machine-readable spec — the four operations and six session schemas were already on this branch; this PR only removes a stale reference to a `GET /ai_assistant/sessions/{id}` the design does not provide |
| `docs/ref/modules/setup/api-reference.md` | published mdBook reference — new **AI Assistant Sessions** section, all four routes with bodies, examples and status codes |
| `docs/ref/security/access-control.md` | rewrites the `wazuh_ai_assistant` section, which documented the unscoped `write` grant as if it were intended |
| `docs/dev/plugins/setup.md` | replaces the sentence claiming the plugin "exposes no administrative API" over sessions; adds the route table and the three data-stream constraints |
| `docs/ref/description.md`, `docs/ref/modules/setup/index.md` | both said access is restricted by DLS without qualifying which half |

`mdbook build` is clean and the new cross-links resolve to real anchors.

</details>

### Tests Introduced

46 new tests. `./gradlew :wazuh-indexer-setup:build` passes Checkstyle, forbidden-APIs and Javadoc.

| Suite | Tests | Scope |
| --- | --- | --- |
| `AuthenticatedUserTests` | 6 | transient parsing, the `_shared` fallback, a user literally named `_shared` |
| `AiAssistantSessionsIndexTests` | 3 | `version` token round-trip; undecodable tokens treated as absent |
| `TransportPutAiAssistantSessionActionTests` | 20 | stamping, the field allowlist, carry-over rules, `updated_at` on rename vs replace, the 500-session cap, `404`/`409` |
| `AiAssistantSessionsIT` | 17 | all four routes end to end on a real cluster, asserting the **stored document**, not just the response |

<details><summary>What the automated suites cannot cover, and two assertions that look redundant</summary>

The `integTest` cluster runs with security disabled, so it has exactly one identity and no
`_opendistro_security_user_info` transient — every session there is stamped `_shared`. That
exercises the whole stamp-and-scope path but cannot demonstrate the impersonation refusal, which
needs a second user. That half is covered by asserting on the document the transport action hands to
the index accessor, plus the live run above. `AiAssistantSettingsAdminIT` documents the same split
for the same reason.

Two integration assertions look redundant and are not: the stale-version test checks the stored
transcript *after* the `409`, because a status code alone would not catch a conflict that retried
and clobbered; and the rename test asserts `updated_at` is unchanged, which is the guard against a
rename reordering the session list.

**Note on `spotlessCheck`.** It fails on a pre-existing Javadoc-reflow violation in
`plugins/setup/src/test/java/com/wazuh/setup/AiAssistantSettingsAdminIT.java`, present on the base
commit and in a file this PR does not touch. Every file added or changed here is Spotless-clean.
`spotlessApply` was deliberately not run, because it reformats that unrelated file and bumps
copyright years elsewhere.

</details>

## Review Checklist

- [x] Code changes reviewed
- [ ] No unresolved dependencies with other issues — **this PR does not close the issue on its own.**
      The `wazuh-indexer` role change (step 3) does, gated on the Dashboard migrating its writes.
- [x] Correct labels applied (`no-changelog`)


MauriFont (changes requested):
**Setup.** Granted `cluster:admin/ai_assistant/session/write` to `wazuh_ai_assistant` and
pushed it with `securityadmin.sh` (the role is `reserved`, so the security API refuses to
patch it). Created a plain internal user `probe` with no backend roles — it gets
`wazuh_ai_assistant` automatically from the `users: ["*"]` mapping.

**Result — it works.** Same forged request as the issue's PoC, sent to the new endpoint:

```
POST /_plugins/_setup/ai_assistant/sessions
{"user":"admin","title":"inyected","messages":[],
"@timestamp":"2026-09-10T09:00:00Z","created_at":"2026-09-10T09:00:00Z",
"updated_at":"2026-09-10T09:00:00Z"}
```

Stored document has `"user": "probe"`, not `"admin"`, and the timestamps are server-stamped rather than the ones sent. Also confirmed the old path is unchanged while the `write` block is still in place.

One change I'd like before this goes in.

### `AuthenticatedUser.parse` doesn't handle the escaped separator

`parse()` cuts at the first `|`:

```java
int separator = userInfo.indexOf('|');
String name = separator < 0 ? userInfo : userInfo.substring(0, separator);
```

The transient escapes pipes in the name: Utils.escapePipe writes `|` as `\|`, and the
canonical reader splits on `(?<!\\)\|` and unescapes (User.java:243-249, common-utils).
This does neither.

For a user named `ad|min` the transient is `ad\|min|backend|own_index|`, so the DLS read filter (${user.name}) is `ad|min` while the document gets stamped `ad\`. They never match: the user writes sessions and can never read them back — the same asymmetry this PR closes. Federated identities carry `|` routinely (an OIDC sub like auth0|68f3c1a9d2, an LDAP DN), and since those share a prefix, every such user collapses onto the same owner `auth0\`.

I confirmed OpenSearch accepts an internal user literally named `ad|min` (created via
.../internalusers/ad%7Cmin), so this is reachable without an IdP.

Failing test for the existing suite:

```
public void testParseHandlesAnEscapedPipeInTheName() {
    assertEquals("ad|min", AuthenticatedUser.parse("ad\\|min|backend|own_index|"));
}
```

Returns `ad\`. testParseTakesTheNameBeforeTheFirstSeparator currently asserts the naive behaviour, so the case is codified rather than missed.

Fix is the same split-and-unescape, or rejecting a name with an unescaped |.

felipegonzalezmv:
Good catch — fixed in https://github.com/wazuh/wazuh-indexer-plugins/pull/1554/commits/cef25307c8c999c90398c44cac54ebd3624678ee: `parse` now splits on unescaped separators and unescapes the name, mirroring `User#parse`; your test case is in the suite along with the `auth0|*` collapse.

Jorgesnchz:
Tests are broken: https://github.com/wazuh/wazuh-indexer-plugins/actions/runs/34502424177/job/102957531670?pr=1554

Local check worked:

``` console
> Task :yamlRestTest NO-SOURCE
> Task :check
> Task :build

BUILD SUCCESSFUL in 3m 57s
53 actionable tasks: 38 executed, 15 up-to-date
13:51:41: Execution finished 'build'.
```

It looks like a cache error

docs/dev/plugins/setup.md
Jorgesnchz, line 1: The role `plugin:wazuh/ai_assistant/session/write` is mentioned, but there is no wazuh-indexer companion PR with the creation of the role

plugins/setup/openapi.yml
```diff
@@ -563,6 +881,226 @@ components:
...
 
+    AiAssistantSessionMessage:
+      type: object
+      description: |
+        One turn of a session.
+
+        **Informative only — this is not a validated contract.** The index maps `messages` as
+        `{"type": "object", "enabled": false}`: the array is stored in `_source` and returned
+        verbatim, but never indexed and never schema-checked. This API treats each entry as an
+        opaque object and validates only the array's length and the document's overall size, so a
```
Jorgesnchz: Looking at the code it looks like the `validates only the array's length and the document's overall size` isn't implemented, it now looks like we have a cap to what can be sent, but we only have the default 100MB one

Jorgesnchz (changes requested):
Small changes needed

docs/dev/plugins/setup.md
felipegonzalezmv: Good catch — there was none. Opened it now: wazuh/wazuh-indexer#1906.

It adds the `plugin:wazuh/ai_assistant/session/write` action group and grants it to `wazuh_ai_assistant`, which is what these endpoints are gated on — without it they answer `403` and the fix here is unreachable.

It deliberately **leaves the index-level `write` block in place**. Removing that block is the step that actually closes #6111, and it has to be gated on the Dashboard having moved its three write calls in `conversation-store.ts` onto the endpoints; landing it earlier means sessions stop persisting, silently. So the sequence is: this PR (endpoints, inert) → #1906 (permission, additive) → Dashboard migration → remove the `write` block.

plugins/setup/openapi.yml
felipegonzalezmv: You're right, and it mattered more than a wording slip: the message cap bounds how many turns a session holds but says nothing about how large each one is, so the only real ceiling was `http.max_content_length` — 100MB by default. That made the 500-session per-owner cap meaningless as a storage bound (500 × 100MB ≈ 50GB per user), on a data stream whose transcripts land in `_source` and are never queried into.

Rather than delete the claim, I made it true — 50a96081:

```java
private static final int MAX_PAYLOAD_BYTES = 5 * 1024 * 1024;
```

Enforced on all three write routes. Two details worth noting:

- Checked against the **raw payload before parsing**, so an oversized body is rejected without being materialised into a `Map` first.
- Measured in **bytes of UTF-8, not characters** — otherwise a multi-byte payload slips past it (`€` is one char but three bytes). There's a test for exactly that.

5 MiB is roughly two orders of magnitude above a realistic chat transcript and two below the platform default. Happy to move the number if you'd rather have it tighter or configurable.

Spec, mdBook reference and dev docs all updated to state the real limit, and the `400` descriptions on all three routes now mention it.

MauriFont (dismissed):
Re-tested on a security-enabled cluster with the current head. My change request is satisfied, and the two changes that landed after my first review are both exercised live now.

**The escaped separator — the thing I asked for.** `cef25307` is not just unit-tested any more. On a real cluster, an internal user literally named `ad|min` creates a session through the endpoint and then **reads it back through the DLS filter**:

```
POST /_plugins/_setup/ai_assistant/sessions        as 'ad|min'   -> 200
GET  /wazuh-ai-assistant-sessions/_search          as 'ad|min'   -> total 1, "user": "ad|min"
```

That is the full loop the truncation would have broken: the stamped owner and the `${user.name}` the read filter substitutes are the same string. Update, rename and delete all resolve as that user too, so the ownership filter agrees with the stamp. Before the fix this user would have written sessions they could never read.

**The 5 MiB cap** (`50a96081`), which landed after the evidence block in the description was produced, so nobody had run it against a cluster:

```
POST /_plugins/_setup/ai_assistant/sessions   ~5.3 MB body
  -> 400 {"message":"Session payload must be 5 MiB or smaller.","status":400}
```

**One thing this PR claims that its own evidence does not actually demonstrate, and that I can now confirm.** Both the description here and in wazuh/wazuh-indexer#1906 show cross-user isolation with a `term` filter on `user`, which does not rule out a reader seeing everything. An unfiltered `match_all` as `admin` returns only `admin`'s own sessions: `all_access` does **not** override the role's DLS, because `plugins.security.dfm_empty_overrides_all` defaults to `false`. So "no impersonation and no administrative override" holds at the index level too, not only at the endpoint. Worth knowing that it rests on that setting — flipping it to `true` would expose every user's conversations to `all_access`.

**Two things found while testing that this PR does not introduce:**

- After an account is deleted, its sessions are unreachable by every authenticated account, `admin` included, for the reason above. The only way in is the admin certificate. There is no operator path to purge a departed user's conversations without filesystem access to a node.
- Recreating an account with a previously used name inherits the earlier holder's conversations — readable, editable and deletable. Ownership is keyed on a mutable, reusable name rather than a stable identifier, which is the same root the original report has. The `SHARED_OWNER` javadoc already notes this class of problem for the `_shared` sentinel; the general case is not covered.

Both would be well served by the admin-gated purge-by-owner endpoint this plugin is already positioned to expose, since it writes with its own privileges and can see what DLS hides.

LGTM.

AlexRuiz7 (changes requested):
Tests fail

ferooo23:
The `gradle-check (content-manager)` failure is not caused by this PR. It is a pre-existing flake in the setup plugin's integration tests, and fixing it needs the same one change in four files under `plugins/setup/src/test/java/com/wazuh/setup/`:

- `DataStreamsIT.java`
- `SetupPluginIT.java`
- `AIAssistantIndicesIT.java`
- `ThreatIntelIndicesIT.java`

```diff
 @After
 public void clearFieldData() throws IOException {
-    Request request = new Request("POST", "/_cache/clear");
+    Request request = new Request("POST", "/wazuh-*,.wazuh-*/_cache/clear");
     request.addParameter("fielddata", "true");
+    request.addParameter("expand_wildcards", "open,hidden");
     client().performRequest(request);
 }
```

**What is going on.** The `@After` in those suites clears the cache with `POST /_cache/clear`, and since no index is specified, OpenSearch applies it to every index in the cluster, system indices included. One of them, `.opendistro-job-scheduler-lock`, does not exist when the cluster starts: ISM creates it around 70 seconds into the cluster's life, and from that moment on it is among the indices the clear reaches.

When a request touches a system index, OpenSearch still serves it (200, 50 shards, 0 failed) but attaches a deprecation `Warning` header. That is where it dies: the test framework's REST client runs in strict mode, so any undeclared warning is turned into a `WarningFailureException`. Since it comes out of an `@After`, JUnit marks the test as failed even though the test body itself passed.

The same `@After` is copied in all four suites, so scoping only `DataStreamsIT` does not fix it , the failure simply moves to the next suite. Verified locally: it reappeared in `SetupPluginIT`.

With this change the request no longer resolves any system index, so no warning is attached and the tests should pass. Verified on a branch off this one, with both gradle-check jobs green:
https://github.com/wazuh/wazuh-indexer-plugins/actions/runs/35213801049

AlexRuiz7 (approved):

Issue #1539, opened 2026-09-04
All three examples the API reference publishes for the content validation endpoint are rejected by it


Rebits:
## Summary

`public-api.yaml` describes the body of `POST /content/validate/resource` with a `resource` member typed only as `object`, with no properties and no schema, so the three worked examples printed beside it are the entire specification of what a resource looks like. All three are rejected by the endpoint. The KVDB example is missing `metadata.title`, and both integration examples are missing `metadata.title` and use a category value the validator does not accept. The error returned names "a valid name", which is not the name of the member that is missing, so the message does not lead the reader to the fix either.

## The schema says nothing

```yaml
    ContentValidate_Request:
      type: object
      required:
      - type
      - resource
      properties:
        type:
          type: string
          enum:
          - decoder
          - filter
          - integration
          - kvdb
        resource:
          type: object
          description: JSON object of the resource to validate
```

`resource` has no properties, no required members and no `$ref`. Everything a reader can learn about the shape comes from the `examples:` block below it.

## The three examples, submitted verbatim

```
   kvdb_validate                        -> 400  {"status":"ERROR","error":"Failed to validate resource of type
                                                 'kvdb': KVDB JSON must have a valid name"}
   integration_validate_with_parent     -> 400  {"status":"ERROR","error":"Failed to validate resource of type
                                                 'integration': Integration JSON must have a valid name"}
   integration_validate_no_parent       -> 400  ... same
```

## What they get wrong

**The title lives under `metadata`.** The examples put it at the top level. Adding `metadata.title` to the published KVDB example, changing nothing else, makes it pass:

```
   such as lo publica the manual            -> 400  KVDB JSON must have a valid name
   anadiendo metadata.title                 -> OK
   only metadata.title, without title loose    -> OK
```

The deployed content agrees with the endpoint and not with the reference:

```
=== a integration real desplegada
   keys: ['category', 'decoders', 'enabled', 'id', 'kvdbs', 'metadata']
   {"id": "7afa379b-...", "metadata": {"title": "apache-http"}, "category": "applications", "enabled": true}

=== a kvdb real desplegado
   keys: ['content', 'enabled', 'id', 'metadata']
   {"id": "add4b6ba-...", "metadata": {"title": "anthropic-event-types"}, "enabled": true}
```

**The category is a closed lowercase vocabulary.** Both integration examples use `Security`. The validator refuses it and accepts the lowercase form:

```
[...]

PR #1555, opened 2026-09-08
Document logsource.product divergence

AlexRuiz7:
## Description

This PR documents the divergence between Sigma's `logsource.product` and its use in Wazuh.

Also fixes a bug where rules were falling back to the "linux" integration by default.

Closes wazuh/internal-devel-requests#6208

## Proposed Changes

`logsource.product` is not a Sigma log-source declaration in 5.0. It is the routing key that maps a rule to its Security Analytics log type, and a log type in 5.0 is a Wazuh integration. Three links make it load-bearing:

1. Each integration is registered in Security Analytics as a log type (`WTransportIndexIntegrationAction`).
2. A rule's compiled query is filed into `.opensearch-sap-{product}-detectors-queries`, keyed by `logsource.product`.
3. Each integration's detector evaluates only the rules stored under its own log type.

So a rule whose `product` does not name the integration that lists it is stored where that integration's detector never reads it, and it matches nothing. The reported catalogue is not violating the Sigma schema — it is satisfying an invariant the indexer enforces at the write path. The docs said otherwise, which is what produced the report.

**Documentation**

- [`ruleset-management/rules.md`](docs/ref/modules/ruleset-management/rules.md) — the Log source section opened with "does not affect detection matching directly" and then, a few lines below, required `product` to match the owning integration's title. That claim is now scoped to `category`, `service` and `definition`, and `### product` describes the routing behaviour, the three links above, and the consequence of a wrong value. Added that platform-looking values (`windows`, `linux`, `apache-http`) are valid only because those integrations are named after platforms — not a second, platform-based meaning for the field. The Sigma reference link now carries the portability caveat: an upstream rule needs its `product` rewritten before ingestion, and an exported rule carries a name other Sigma tooling does not recognise. The starting example uses `product: wazuh-generic-1` instead of `product: python`, which was not a valid value.
- [`ruleset-management/architecture.md`](docs/ref/modules/ruleset-management/architecture.md) — the `.opensearch-sap-log-types-config` row said "Integrations"; it now says log types, and why every integration being one is what makes `logsource.product` name an integration.

**Bug fix — the "linux" default**

The invariant was enforced asymmetrically. [`TransportCreateRuleAction`](plugins/content-manager/src/main/java/com/wazuh/contentmanager/transport/TransportCreateRuleAction.java#L156-L168) returns 400 when a user-created rule's `logsource.product` disagrees with its integration, but the CTI sync path validated nothing — [`ContentIndex.extractProduct()`](plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/index/ContentIndex.java#L684) read the field, fell back to `logsource.category`, and otherwise defaulted to the string `"linux"`:

```java
String product = "linux";
if (ruleNode.has(KEY_LOGSOURCE)) { ... }
```

A content build shipping a rule with a missing or renamed `logsource.product` therefore filed it silently under the `linux` log type, or under a category name like `process_creation` — no error, no finding ever produced, sync reporting success. Since this invariant is load-bearing and lives only in the content, that is a defect worth closing on the sync path:

- **`ContentIndex.extractProduct()`** returns `null` when the field is absent. No substitute value: both the category fallback and the `"linux"` default are gone, because either one files the rule under a log type that is not its integration's.
- **`SecurityAnalyticsServiceImpl.upsertRule()`** rejects a rule that declares no product. Every rejection on that method now notifies the listener — the pre-existing early returns were silent, and callers count these requests down against a `CountDownLatch`, so a malformed rule made the caller wait out its whole timeout and report *that* instead of the rule. Also corrects `!doc.has(KEY_METADATA) && !doc.get(KEY_METADATA).isObject()` to `||`: as written it dereferenced a null node when `metadata` was absent, and short-circuited past a non-object `metadata` when it was present.
- **`ConsumerRulesetService.syncRules()`** now receives the integration documents already read for this pass and cross-checks each rule's `product` against the `metadata.title` of the integration that lists it. A mismatch is logged and held back rather than sent, and leaves `rules` in the pending phases so the next sync retries. Integration docs are read once and shared across the integrations, rules and detectors phases. If that read did not complete, rules it does not cover are sent unchecked rather than held back, so a failed integrations read cannot stall the rule sync.

**Out of scope**

The issue's proposed fix — setting `product` to `windows`, `linux`, `apache` — is not adopted. The create-rule endpoint would reject those values, and on the sync path they would misroute every rule to a log type that does not exist. The section arguing that `product` disagrees with the events its rules match does not hold either: rule-to-event association does not run through `logsource.product` at all, since `DetectorFactory` receives an explicit rule-id list and explicit input indices. The 3,902 findings cited are evidence that the routing works.

The portability cost is real and is the part of the issue that survives, but decoupling is a design decision for @wazuh/devel-xdrsiem-indexer rather than a defect fix. It is not cheap to do incidentally: `cti-rules-mappings.json` has no `integration` field on the rule document — membership is recorded only on the integration's `rules` list, which is precisely why `logsource.product` became the routing key. Decoupling means adding an explicit envelope field (`document.integration`) and migrating, in exchange for Sigma rules that round-trip. This PR documents the current behaviour so the trade can be discussed against an accurate description of it.

### Results and Evidence

Full `content-manager` unit suite, on this branch:

```
$ ./gradlew :test
BUILD SUCCESSFUL

746 tests, 0 failures, 0 errors
```

`ConsumerRulesetServiceTests` — 21 tests, 0 failures, including the new mismatch case:

```
tests="21" skipped="0" failures="0" errors="0"
testOnSyncComplete_ruleProductMismatchesIntegration_isNotSent  PASSED
```

Both call sites of `extractProduct()` are covered by the change; there are no others:

```
SecurityAnalyticsServiceImpl.java:190:  String product = ContentIndex.extractProduct(doc);
ConsumerRulesetService.java:527:        String product = ContentIndex.extractProduct(doc);
```

Log lines a mismatched or product-less rule now produces, replacing the previous silence:

```
Rule [rule-mismatch] declares 'logsource.product' [windows] but belongs to integration
[apache-http]. It is not sent to Security Analytics, because it would be filed under a log
type that integration's detector does not read.

Rule [rule-id] declares no 'logsource.product'. It is not sent to Security Analytics,
because the field is the log type the compiled query is filed under.
```

### Artifacts Affected

- `wazuh-indexer-content-manager` plugin.
- Developer documentation under `docs/ref/modules/ruleset-management/`.

No packaging, executable or index-mapping changes.

### Configuration Changes

None. No new settings, no changed defaults.

Behavioural note for content builds rather than for configuration: a CTI rule that omits `logsource.product`, or declares one that disagrees with the integration listing it, is no longer synced to Security Analytics under a substituted log type. It is now held back and logged at `ERROR`. Such a rule previously produced no findings either — the change is that the failure is now visible and the phase is retried, instead of the sync reporting success.

### Documentation Updates

- `docs/ref/modules/ruleset-management/rules.md` — Starting example, top-level field list, Log source section intro, `product`, `category`, `service`, and the Sigma reference note.
- `docs/ref/modules/ruleset-management/architecture.md` — the `.opensearch-sap-log-types-config` row in the index table.

`ref/modules/content-manager/api.md` and `openapi.yml` already stated the constraint correctly and needed no change.

The `.claude/skills/docs-review/` skill and its style guide are updated with the two review checks that would have caught this: verifying what a documented field *does* rather than only that its stated constraint is right, and reading a section intro's blanket claim against every subsection beneath it. Every individual sentence on the old page was defensible; only asking "what breaks if this value is wrong?" surfaced the problem.

### Tests Introduced

`ConsumerRulesetServiceTests.testOnSyncComplete_ruleProductMismatchesIntegration_isNotSent` — an integration titled `apache-http` lists two rules, one declaring `product: apache-http` and one declaring `product: windows`. Asserts that only the matching rule reaches `upsertRule`, and that `rules` is left in the consumer document's pending phases so the next sync retries. `windows` is deliberately a real integration name, so the test covers a plausible mismatch rather than a nonsense value.

## Review Checklist

- [ ] Code changes reviewed


Jorgesnchz (approved):
LGTM

AI is flagging some changes, but they will be solved in https://github.com/wazuh/wazuh-indexer-plugins/issues/1539. The only thing concerning for me is that there will be conflicts with the other PR

PR #1557, opened 2026-09-09
Document the removal of the OpenSearch demo users

MauriFont:
## Description

Companion documentation for wazuh/wazuh-indexer#1899, which stops the packages from shipping the OpenSearch Security demo users and the role mappings that lead to their roles. No page named those accounts, so nothing was factually wrong — what was missing was the new behavior:

- **`docs/guide/migration/index.md`** — the "Security migration" intro now states that 5.x no longer ships the upstream demo entries, and step 2 carries a warning: the removed mappings keyed on backend role *names*, so a deployment with an LDAP/AD group, JWT claim or certificate attribute called `kibanauser`, `readall`, `logstash` or `snapshotrestore` silently loses those privileges — users authenticate as before and nothing reports the cause.
- **`docs/ref/security/defining-users-and-roles.md`** — a warning not to map users to the built-in `kibana_user` role. Its name reads as plain dashboard access, but it grants `delete`/`index`/`manage` over `.kibana*`, and with multi-tenancy disabled a user holding it can permanently delete any shipped index pattern or silently repoint it. The role is static and cannot be removed or weakened from a configuration file, so the warning is the only remaining defense.

Closes wazuh/internal-devel-requests#6113

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/docs/ref/security/defining-users-and-roles.md
+++ b/docs/ref/security/defining-users-and-roles.md
@@ -30,6 +30,18 @@ Follow these steps:
     * Optionally, configure [**Document-level security (DLS)**](https://docs.opensearch.org/3.6/security/access-control/index/) or [**Field-level security (FLS)**](https://docs.opensearch.org/3.6/security/access-control/field-level-security/).
 6. Click **Create** to save the role.
 
+<div class="warning">
+
+**Do not map users to the built-in `kibana_user` role.**
+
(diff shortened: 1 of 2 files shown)


docs/ref/security/defining-users-and-roles.md
Jorgesnchz, line 1: I'd either change this to make it clearer or delete the block, IMO is confusing

Jorgesnchz (approved):
I'll leave the comment and approve, so @AlexRuiz7 can decide if we should apply the change and delete the comment or merge with it, to avoid having to do a change that is only my opinion

AlexRuiz7:
I do not think we need documentation to justify why we decided not to ship OpenSearch demo users. Closing the PR.

PR #1559, opened 2026-09-09
 Adopt `match_only_text` and `wildcard` in the WCS

### LINKED ISSUE #1529: `message` is mapped as `keyword` instead of ECS's `match_only_text`, so log text cannot be searched by word
**Issue title:** [BUG] `message` is mapped as `keyword` instead of ECS's `match_only_text`, so log text cannot be searched by word

**Describe the bug**

ECS defines `message` as `match_only_text`, indexed, described as "Log message optimized for viewing in a log viewer". The events stream template declares it as a `keyword` with `ignore_above: 1024` instead.

Two consequences follow: word-level search over log text does not work at all, and any message longer than 1024 characters is not indexed even for exact match, while still being present in `_source`.

**To Reproduce**

```bash
# the mapping
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events-v5-system-activity/_mapping/field/message,event.original'

# ingest a decodable line via an agent localfile, then:
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events*/_count' -H 'Content-Type: application/json' \
  -d '{"query":{"match":{"message":"Failed password"}}}'          # 0
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events*/_count' -H 'Content-Type: application/json' \
  -d '{"query":{"term":{"process.name":"sshd"}}}'                 # the structured field works
```

**Expected behavior**

`message` mapped as ECS specifies, `match_only_text`, so that word-level search over log content works. That type is already available and already in use in this very template, which is what makes this look like an oversight rather than a constraint:

```
dynamic templates in the events stream template: 2299
of those using match_only_text:                     1     ->  file.diff
message:  declared in explicit properties as keyword, ignore_above 1024
```

So the WCS generator can emit `match_only_text`, and does, for exactly one field. `message`, the one field ECS defines that way, is hand-declared as a keyword instead.

**Plugins**

**Host/Environment**

**Additional context**

**Preconditions and considerations**

Stock single-node deployment, events stream template exactly as shipped, no mapping changes. The measurement needs events that actually decoded, so the ten reference events were produced by feeding `sshd` authentication lines through an agent `localfile`.

One trap is worth stating because it invalidated an earlier attempt at this finding: padding used to test `ignore_above` must go **inside** a field the decoder extracts, not appended to the end of the line. Appending it after `ssh2` breaks the `system-auth` decoder's anchor, so nothing decodes at all and the result looks like a length limit that is really a decoding failure. Both false leads that came out of this are recorded at the bottom of this file.

**Observed**

The mapping, hand-declared in the template's explicit `properties`:

```json
"message": {"type": "keyword", "ignore_above": 1024}
```

Measured against ten known ingested events, the structured fields the decoder extracted all work and the text does not:

```
term  process.name = sshd            -> 10   OK
term  event.action = auth-failure    ->  9   OK
term  source.ip = 198.51.100.9       ->  3   OK
term  user.name = oracle             ->  3   OK

match message 'Failed password'      ->  0   no word-level search
query_string 'oracle'                ->  0   no free-text search
query_string message:Failed          ->  0
```

And the `ignore_above` half, measured by feeding lines of increasing length with the padding inside a decoded field so the decoder's anchor stayed intact:

```
user.name=  990 chars  message= 1054 chars   searchable by that field
user.name= 1190 chars  message= 1254 chars   NOT searchable by that field
user.name= 1490 chars  message= 1554 chars   NOT searchable by that field
```

Nothing is dropped for length: a 1554-character line still produces a complete event, retrievable by `wazuh.protocol.location`, with the full value in `_source`. It is the field that is not indexed, not the document that is missing, so the data looks fine once you have found it by other means. `user.name` carries the same `keyword` with `ignore_above: 1024`, so the rule applies to every field mapped that way.

### Root cause, verified in source

The mapping is hand-declared in the events stream template, `wazuh-indexer-plugins` branch `5.0.0`, `plugins/setup/src/main/resources/templates/streams/events.json`:

```json
"message": {"type": "keyword", "ignore_above": 1024}
```

Read programmatically from that file, together with the dynamic templates in the same document:

```
message                                     -> {"ignore_above": 1024, "type": "keyword"}
dynamic_templates in the same template      -> 2299
of those emitting match_only_text           -> 1   (wcs_file_diff)
```

So the generator that produces this template *can* emit `match_only_text`, and does, for exactly one field: `file.diff`. `message`, the one field ECS defines as `match_only_text`, is the hand-written exception.

Two distinct consequences follow from the one declaration, and they need separate tests:

1. `type: keyword` means the field is indexed as a single token, so no word-level or free-text query can match inside it.
2. `ignore_above: 1024` means any message longer than 1024 characters is not indexed at all, not even for exact match, while still being returned in `_source`. It therefore looks present and is unsearchable.

**How to reproduce, minimal and self-contained**, without needing an agent:

```bash
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events-v5-*/_mapping/field/message?pretty' | grep -A3 '"message"'
# "type" : "keyword", "ignore_above" : 1024

# a word-level query over a message known to be present returns nothing
curl -sk -u admin:admin 'https://localhost:9200/wazuh-events-v5-*/_search?size=0' \
  -H 'Content-Type: application/json' -d '{"query":{"match":{"message":"Failed password"}}}'
# hits.total.value = 0, while the same documents are returned by a term query on a decoded field
```

### Worse than the mapping alone: the fallback field cannot be queried, and half the events have no searchable copy

Two further facts, found by sweeping every mapped leaf field in the events and findings indices for ones that no document populates.

**`event.original` is not indexed at all.** It is the field that always carries the raw log line, and its mapping disables both search and aggregation:

```
event.original: {"type": "keyword", "index": false, "doc_values": false}
```

Every attempt to query it fails outright rather than returning nothing:

```
match       on event.original  ->  search_phase_execution_exception
match_phrase                   ->  search_phase_execution_exception
wildcard                       ->  search_phase_execution_exception
term        (exact full value) ->  search_phase_execution_exception
```

So the copy of the log line that is always present can be retrieved once a document is found, and can never be used to find one.

**`message` is populated for some integrations and not others.** The sweep found it empty in every document of the security stream, and the per-index counts show why:

```
.ds-wazuh-events-v5-system-activity-000001    message populated in  8351 documents   (linux integration)
.ds-wazuh-events-v5-security-000001           message populated in     0 of 868       (wazuh-rootcheck)
wazuh-findings-v5-security                    message populated in     0
```

For a rootcheck event there is therefore **no searchable copy of the text at all**: `message` is absent and `event.original` cannot be queried. The 868 security events on this deployment are retrievable but not findable by content.

**And free-text search returns nothing.** A `query_string` for a word that appears in 868 documents finds none of them, because the only field holding it is unindexed:

```
{"query":{"query_string":{"query":"hidden"}}}    ->  0
```

while `event.original` in those same documents reads `Files hidden inside directory '/usr/sbin'. ...`

Where `message` *is* populated, the keyword mapping behaves exactly as this finding describes:

```
term     on the complete exact value   ->  1        works
match    on a single word              ->  0        does not
wildcard *10.5.5.5*                    ->  2630     works, at the cost of a full scan
```

### Impact

An analyst cannot ask "every event mentioning this string", which is a normal request during an investigation and the reason Discover exists. Only fields a decoder happened to extract are queryable, so anything a decoder did not model is unreachable even though it was ingested and stored.

### Tasks

- [ ] Map `message` as `match_only_text` in the events stream template, as ECS defines it and as the WCS generator already emits for `file.diff`.
- [ ] Check the other hand-declared `keyword` fields in the same explicit `properties` block for the same deviation, rather than fixing only the one field this issue names.
- [ ] Decide what should happen to messages over `ignore_above`, which today are stored in `_source` and not indexed at all, so they are retrievable only by another field.
- [ ] Add a test that ingests a decodable line and asserts a word-level `match` on `message` returns it.
- [ ] Leave `event.original` alone: its `index: false` is ECS-conformant and deliberate, see the retraction below.

### Retracted from an earlier version of this finding

This file previously also reported `event.original` as a defect, on the grounds that `index: false` and `doc_values: false` make the raw line unsearchable. **That part is withdrawn: it is ECS-conformant and deliberate.**

ECS itself declares:

```
event.original    type=keyword    Indexed=false    "Raw text message of entire event."
message           type=match_only_text   Indexed=true
```

And Wazuh declares it explicitly in the WCS, as one of only 13 fields out of 2299 with `index: false`, the rest being `x509.public_key_exponent` variants and `gen_ai.agent.description`:

```json
"wcs_event_original": {"path_match": "event.original",
  "mapping": {"doc_values": false, "index": false, "type": "keyword"}}
```

Present identically on `5.0.0` and `main`. So `event.original` is not searchable by design, matching the schema, and only the `message` mapping is a deviation. Checking the spec before writing the finding would have caught this; it is the same mistake that produced the retraction in [016](not-for-filing/not-a-defect/NON_BUG_DETECTED_016-RETRACTED-unclassified-events-not-indexed.md).

### Two false leads, also mine

Recorded because both nearly became findings:

- **"Long lines produce no event."** They do. My first attempt appended padding *after* `ssh2`, where the `system-auth` decoder anchors, so nothing decoded for a reason unrelated to length. Even the 200-character line failed. The tell was that the short line should have worked and did not.
- **"None of the padded lines arrived."** They did, with an empty `user.name`, and I had queried that field and read the empty result as absence. The same mistake appears in [016](not-for-filing/not-a-defect/NON_BUG_DETECTED_016-RETRACTED-unclassified-events-not-indexed.md): querying a field that the defect itself empties is a reliable way to convince yourself there is no data.

Jorgesnchz:
## Description

The WCS sanitizer rewrote four ECS field types to equivalents because the Initial release of the WCS had no support for those type of fields

| ECS type | live 5.0.0 verdict | decision |
|---|---|---|
| `match_only_text` | **ACCEPTED** | **adopt** |
| `wildcard` | **ACCEPTED** | **adopt** |
| `constant_keyword` | **REJECTED** — `mapper_parsing_exception: Field [type] is missing required parameter [value]` | keep remapping to `keyword` |
| `flattened` | **REJECTED** — `No handler for type [flattened]` | keep remapping to `flat_object` |

Closes https://github.com/wazuh/wazuh-indexer-plugins/issues/1529

## Proposed Changes

Modify the list of fields to remap in `wcs/generator/images/schema_sanitizer.py`:

```diff
 TYPES_TO_REMAP = {
     'constant_keyword': 'keyword',
-    'wildcard': 'keyword',
-    'match_only_text': 'keyword',
     'flattened': 'flat_object',
 }
```

Generate again the files using the generator

### Results and Evidence

Every type transition across all 23 templates:
``` console
   keyword -> wildcard              53
   keyword -> match_only_text       24
   TOTAL field mappings changed     77
```

All 40 templates install and work on a live 5.0.0 node.

### Artifacts Affected

Index templates

### Configuration Changes

NA

### Documentation Updates

NA

### Tests Introduced

NA

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/plugins/setup/src/main/resources/templates/states/agent-config.json
+++ b/plugins/setup/src/main/resources/templates/states/agent-config.json
@@ -23,8 +23,7 @@
       "dynamic": "true",
       "properties": {
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "state": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/agent-stats.json
+++ b/plugins/setup/src/main/resources/templates/states/agent-stats.json
@@ -21,8 +21,7 @@
       "dynamic": "true",
       "properties": {
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "state": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/fim-files.json
+++ b/plugins/setup/src/main/resources/templates/states/fim-files.json
@@ -127,8 +127,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "state": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/fim-registry-keys.json
+++ b/plugins/setup/src/main/resources/templates/states/fim-registry-keys.json
@@ -53,8 +53,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "registry": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/fim-registry-values.json
+++ b/plugins/setup/src/main/resources/templates/states/fim-registry-values.json
@@ -53,8 +53,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "registry": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-browser-extensions.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-browser-extensions.json
@@ -84,8 +84,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "package": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-groups.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-groups.json
@@ -67,8 +67,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "state": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-hardware.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-hardware.json
@@ -83,8 +83,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "state": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-hotfixes.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-hotfixes.json
@@ -43,8 +43,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "package": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-interfaces.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-interfaces.json
@@ -110,8 +110,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "state": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-networks.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-networks.json
@@ -51,8 +51,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "network": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-packages.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-packages.json
@@ -55,8 +55,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "package": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-ports.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-ports.json
@@ -99,8 +99,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "network": {
           "properties": {
--- a/plugins/setup/src/main/resources/templates/states/inventory-processes.json
+++ b/plugins/setup/src/main/resources/templates/states/inventory-processes.json
@@ -52,8 +52,7 @@
           }
         },
         "message": {
-          "ignore_above": 1024,
-          "type": "keyword"
+          "type": "match_only_text"
         },
         "process": {
           "properties": {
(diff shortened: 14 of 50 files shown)


Jorgesnchz:
I believe that the best way to test this change is to wait for the rest of the affected teams (This change may impact the Engine, the ruleset and the Dashboard. ([ref](https://github.com/wazuh/wazuh-indexer-plugins/issues/1529#issuecomment-5572759651)))

And do a complete test of the flow, but since Indexer since is the first that needs to be done I have applied the changes and we can review them

Also, I think we could see how the performance works for the `wildcard` type fields maybe it makes it works and is not worth for us

AlexRuiz7 (changes requested):
We need to discuss the `constant_keyword` field.

AlexRuiz7 (approved):
While Elastic defines the `value` parameter for the `constant_keyword` field optional, OpenSearch sets it as required.

The `value` holds the constant value of the field, which we do not know at index template generation time.

Given this evidence, I'm approving this pull request.

- https://raw.githubusercontent.com/elastic/ecs/refs/heads/main/generated/elasticsearch/legacy/template.json
- https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/keyword#constant-keyword-params
- https://docs.opensearch.org/latest/mappings/supported-field-types/constant-keyword/#parameters

PR #1565, opened 2026-09-10
Document the threat detector detection gap and detector permissions

### LINKED ISSUE #1531: Shipped detectors are protected from deletion and not from being switched off, and the detections missed while one is off are never recovered
**Issue title:** `[BUG] Shipped detectors are protected from deletion and not from being switched off, and the detections missed while one is off are never recovered`

**Describe the bug**

The 21 detectors a deployment ships with are treated as protected: an explicit guard, above and separate from the permission system, refuses to delete them. That guard exists because losing them would disable detection.

Setting `enabled: false` on the same detector has the same effect and is not guarded. It is allowed for any account whose role carries the Security Analytics write permissions, which on a stock deployment includes an account whose password is published in the image.

**To Reproduce**

On any stock deployment:

```bash
# choose a detector and save su definicion
DET=$(curl -sk -u admin:admin -H 'Content-Type: application/json' \
  'https://localhost:9200/_plugins/_security_analytics/detectors/_search' \
  -d '{"query":{"match_all":{}},"size":1}' | python3 -c 'import sys,json;print(json.load(sys.stdin)["hits"]["hits"][0]["_id"])')

# deleting it is forbidden, above RBAC
curl -sk -u wazuh-demo:wazuh-demo -X DELETE \
  "https://localhost:9200/_plugins/_security_analytics/detectors/$DET"
# -> 400 "Standard detectors cannot be deleted by users."

# disabling it no lo this
curl -sk -u admin:admin "https://localhost:9200/_plugins/_security_analytics/detectors/$DET" \
  | python3 -c '
import sys,json,datetime
d=json.load(sys.stdin); s=d.get("detector",d)
for k in ("last_update_time","enabled_time"):
    v=s.get(k)
    if isinstance(v,str) and v.endswith("Z"):
        s[k]=int(datetime.datetime.fromisoformat(v.replace("Z","+00:00")).timestamp()*1000)
s["enabled"]=False
json.dump(s,open("/tmp/off.json","w"))'
curl -sk -u wazuh-demo:wazuh-demo -X PUT -H 'Content-Type: application/json' \
  "https://localhost:9200/_plugins/_security_analytics/detectors/$DET" --data-binary @/tmp/off.json
# -> 200, and enabled ends up in false
```

Note the payload detail, because it wastes time otherwise: the detector API returns `last_update_time` and `enabled_time` as ISO strings and rejects them on write, expecting epoch milliseconds. An unconverted payload fails with `illegal_argument_exception: For input string: "2026-08-25T09:13:32.762Z"`, which is a parse error and not an authorisation one.

For the impact half, disable the `linux` detector, write `Accepted password for root from <ip> port 22 ssh2` on an agent, and compare the event and finding counts for that address against the same test with the detector enabled.

**Expected behavior**

- The guard that protects standard detectors from deletion should protect them from being disabled, or from any update that stops them running. If disabling them is a legitimate operation, it belongs to an administrator, not to a role shipped for demonstration.
- `wazuh_demo` should not carry write permissions on the detection configuration. Read access to content and detectors is what a demo role needs.
- A detector changing state should be recorded somewhere an operator will see, since the change is silent and its effect is invisible until someone compares expected findings against actual ones.
- Whether the missed window can be re-evaluated is a larger question and probably out of scope, but the report should say plainly that it cannot be, because that is what makes the gap permanent rather than temporary.

**Plugins**

**Host/Environment**

**Additional context**

**Observed**

The guard on deletion works, and it is clearly deliberate:

```
DELETE /_plugins/_security_analytics/detectors/<id>    (as wazuh-demo)
  -> 400 {"type":"status_exception","reason":"Standard detectors cannot be deleted by users."}
```

For contrast, an account without the permission is stopped earlier and differently, by RBAC:

```
DELETE /_plugins/_security_analytics/detectors/<id>    (as readall)
  -> 403 {"type":"security_exception",
          "reason":"no permissions for [cluster:admin/opensearch/securityanalytics/detector/delete]"}
```

Two different layers, both doing their job. Now the same detector, same account, one field changed:

```
PUT /_plugins/_security_analytics/detectors/<id>       (as wazuh-demo)
    {... "enabled": false ...}
  -> 200
```

and it takes effect:

```
before:   name: auditd | enabled: True
after: name: auditd | enabled: False
```

### What it costs, measured

The `linux` detector was used because this battery had already established exactly what it produces: an sshd "Accepted password for root" line matches rule `94d51c3e` and yields one finding per event.

**With the detector switched off by `wazuh-demo`,** ten such events were written on an agent:

```
events with source.ip 10.99.0.1:  10      <- decoded and stored
findings from those events:          0
```

The events are not lost. They are ingested, decoded with the full ECS field set and stored. They simply produce no detection.

**With the detector on,** ten identical events, differing only in the source address:

```
docs with source.ip 10.99.0.2:  20         <- 10 events + 10 findings
```

**And the gap does not heal.** After re-enabling the detector, the events from the blind period were re-checked for four minutes:

```
   +30s  findings of 10.99.0.1 (blind period) = 0
   +60s  ... = 0
   ...
   +240s ... = 0
```

Still zero. Security Analytics detectors evaluate forward from their schedule; nothing re-runs them over the window they were off. So an attacker who disables a detector, acts, and re-enables it leaves the events in the index and no findings anywhere, and the deployment cannot recover the missed detections after the fact.

### Why this is a defect and not a permission the role is supposed to have

Three things together, none of which is sufficient alone.

**The guard is inconsistent with itself.** Somebody decided the shipped detectors must not be destroyed by users and enforced it above RBAC. Disabling one produces the same operational outcome as deleting it, for the duration, and passes.

**The role that can do it is a demo role with a published password.** `wazuh-demo`'s description in the shipped `internal_users.yml` is "Wazuh user with read access to consumers and threat intelligence content ... plus full access to Security Analytics plugin endpoints", and its hash is documented in the same file as the hash of `wazuh-demo`. So on a stock deployment this is reachable with a credential anyone can read, on a port the shipped compose publishes. `wazuh-admin` holds it too, which is defensible for an administrator; `wazuh-demo` holding it is not.

**Nothing records it.** No alert, no finding, no entry in the manager log. The only trace is the detector's own `enabled` field, and an operator has no reason to look at it. This is the same shape as the findings in this battery where the product reports success while doing nothing, except here it is an attacker choosing when it happens.

### Tasks

- [ ] ~Extend the standard-detector guard to cover disabling, or any update that stops a shipped detector running.~
- [ ] Remove detection-write permissions from `wazuh_demo`. Its own description says it is a demo account.
- [ ] Record detector state changes somewhere visible. Today the change leaves no trace outside the detector document.
- [ ] Decide whether a re-evaluation of a window a detector was disabled for is feasible; if not, say so in the documentation, because operators will assume re-enabling recovers the gap.
- [ ] Add a suite case asserting that a non-administrator cannot change `enabled` on a shipped detector. It is the same request as the delete case that is already guarded.

felipegonzalezmv:
## Description

Standard detectors are protected from deletion but can be switched off, which stops detection for their integration just as effectively. Nothing in the reference said so, nor that the detections missed while a detector is off are never recovered.

Closes #1531, which needs all three of:

| | |
|---|---|
| this PR | documents the detection gap and the permissions |
| wazuh/wazuh-indexer-security-analytics#332 | logs each transition with the account that requested it |
| wazuh/wazuh-indexer#1903 | restricts `wazuh_demo` to read-only access on detectors |

## Proposed Changes

A new **Enabling and disabling detectors** section in the Ruleset Management reference, plus the security pages that follow from the role change.

| Documented | Where |
|---|---|
| A standard detector accepts a change to `enabled` and nothing else, and cannot be deleted — both rejection messages verbatim | `ruleset-management/index.md` |
| The API returns `last_update_time` / `enabled_time` as ISO-8601 but expects epoch milliseconds on write; an unconverted payload fails with `illegal_argument_exception`, a parse error and not an authorisation one | same |
| Who may switch a detector off | same |
| **The detection gap** — no findings while off, the missed interval is never re-evaluated, the events stay searchable, and there is no re-scan or backfill API | same |
| The INFO log entry each transition leaves | same |
| `wazuh_demo` is read-only on detectors; `detector/write` is what switches one on or off | `security/access-control.md`, `security/permissions.md` |
| `cluster:admin/wazuh/securityanalytics/detector/set_enabled`, missing from the permission list | `security/permissions.md` |

The gap claim is what the code does, not a repetition of the issue thread: a detector's `enabled` becomes its composite workflow's `enabled` (`WorkflowService`), and re-enabling a workflow sets `isWorkflowRestarted`, which passes `forceCreateLastRunContext = true` to `MonitorMetadataService.getOrCreateMetadata` (`TransportIndexWorkflowAction`; same logic for standalone doc-level monitors in `TransportIndexMonitorAction`). That rebuilds `lastRunContext` at the shards' current sequence numbers, so the skipped interval is dropped deliberately.

**What this does not do:** it does not make the missed window recoverable, and it changes no behaviour.

### Results and Evidence

Every documented claim was checked against a real cluster or against the source. `mdbook build` renders and the cross-page anchors resolve.

<details><summary>Verified on a cluster</summary>

A `vagrant/5.0` VM (OpenSearch 3.6.0, security plugin enabled) with all three branches deployed via `push.sh --build --install`, against `apache-http`, a real **standard** detector:

| Documented claim | Result |
|---|---|
| standard detectors cannot be deleted | 400 `Standard detectors cannot be deleted by users.` (403 first for a role without `detector/delete`) |
| no update but the `enabled` toggle | 400 `Standard detectors cannot be modified by users. Only enabling or disabling is allowed.` |
| `wazuh_demo` cannot switch one off | 403 on `detector/write`; detector unchanged |
| an administrator still can | 200 |
| each transition is logged | `Detector [7afa379b-…] (apache-http, standard) was disabled by [wazuh-admin]. …` |
| epoch milliseconds on write | `400 illegal_argument_exception: For input string: "2026-09-04T16:03:04.772Z"` |

</details>

<details><summary>Verified by reading the source</summary>

| Claim | Source |
|---|---|
| the toggle requires `detector/write` | `IndexDetectorAction.NAME` |
| `detector/set_enabled` is the Content Manager's path | `WSetDetectorEnabledAction.NAME`, from `TransportUpdateIntegrationAction.syncDetectorEnabledState` |
| times are epoch millis on write | `Detector.parse` → `Instant.ofEpochMilli(xcp.longValue())` |
| re-enabling resets the run context | `TransportIndexWorkflowAction` / `TransportIndexMonitorAction` → `forceCreateLastRunContext` |

**Not empirically tested:** the gap itself — indexing an event while a detector is off and showing no finding is produced for it afterwards. The mechanism is proven in code and stated by the maintainer in #1531, but no event was pushed through the pipeline.

</details>

### Artifacts Affected

None. Documentation only — no plugin, packaging, default configuration, index template or security configuration. `docs/book/` is build output and is not committed.

### Configuration Changes

None in this PR. The `wazuh_demo` change it documents lives in wazuh/wazuh-indexer#1903; existing deployments keep the roles they have until the new `roles.yml` is applied.

### Documentation Updates

- `docs/ref/modules/ruleset-management/index.md` — the new section; the API bullet on detector updates links to it.
- `docs/ref/security/access-control.md` — `wazuh_demo` in the user list and the role breakdown.
- `docs/ref/security/permissions.md` — the role summary, `detector/set_enabled`, and a note on `detector/write`.

No endpoint changed, so `openapi.yml` needs no update. No new page, so `docs/SUMMARY.md` is untouched.

### Tests Introduced

None — no code to test. Verification is the cluster evidence above plus `mdbook build`.

## Review Checklist

- [x] Code changes reviewed
- [ ] Tests cover the new functionality — not applicable, documentation only
- [ ] No unresolved dependencies with other issues — describes behaviour that exists once wazuh/wazuh-indexer-security-analytics#332 and wazuh/wazuh-indexer#1903 merge
- [x] Correct labels applied (e.g., `no-changelog`) — CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -190,6 +190,7 @@
 - [BUG] Resource creation requires `indices:admin/create` on `.wazuh-content-manager-resource-locks` for the calling user [(#1520)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1520)
 - [BUG] `index.query.default_field` lists name fields the index does not have, fields no text term can match and entries with trailing spaces [(#1536)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1536)
 - [BUG] `logtest` reports `contains` matches that never fire in the real ingestion pipeline (case sensitivity) [(#321)](https://github.com/wazuh/wazuh-indexer-security-analytics/issues/321)
+- [BUG] Document that a disabled threat detector produces no findings and that the missed interval is never re-evaluated [(#1531)](https://github.com/wazuh/wazuh-indexer-plugins/issues/1531)
 
 ## Prior versions
 - []()
--- a/docs/ref/modules/ruleset-management/index.md
+++ b/docs/ref/modules/ruleset-management/index.md
@@ -32,6 +32,47 @@ When the restriction is violated, the API returns `400 Bad Request`.
 
 Both limits are dynamic and enforced at the transport layer, applying to all detector creation and update paths, including inter-plugin calls from the Content Manager. Both accept any value from `0` upwards; there is no hard-coded ceiling. See [Configuration](configuration.md) for details.
 
+## Enabling and disabling detectors
+
+A standard detector accepts one user change and one only: switching `enabled` on or off. Every other field is owned by the CTI catalog and is rejected with `400 Bad Request` and `Standard detectors cannot be modified by users. Only enabling or disabling is allowed.` A standard detector also cannot be deleted; the API answers `400 Bad Request` with `Standard detectors cannot be deleted by users.`
+
+```bash
+# Disable a detector (as an administrator)
+curl -sk -u wazuh-admin:<password> -X PUT \
+  -H 'Content-Type: application/json' \
+  "https://localhost:9200/_plugins/_security_analytics/detectors/<detector_id>" \
+  --data-binary @detector-with-enabled-false.json
+```
+
+> **Note:** the detector API returns `last_update_time` and `enabled_time` as ISO-8601 strings but expects epoch milliseconds on write. Sending back an unmodified response body fails with `illegal_argument_exception: For input string: "..."`, which is a parse error, not a permission one.
+
+### Who can switch a detector off
+
+Switching a detector on or off requires `cluster:admin/opensearch/securityanalytics/detector/write`, which among the [default roles](../../security/access-control.md) only `wazuh_admin` holds. `wazuh_demo` and `wazuh_readonly` can read detectors (`detector/get`, `detector/search`) but cannot change their state.
+
+### The detection gap
+
+**A disabled detector produces no findings, and the events missed while it was off are never recovered.** A detector analyses the events of each scheduled run, and a disabled detector is not run at all. Events indexed while it was off therefore fall outside every run, before and after, so re-enabling it resumes detection from that moment forward — it does not go back over the interval that was skipped.
+
+The consequences are worth stating plainly:
+
+- The events themselves are not lost; they are indexed as usual and remain searchable in `wazuh-events-v5-*`.
+- No finding exists for them and none will be created later, so anything that reads findings — the Dashboard, [case management](case-management.md), triggers and alerts on user-created detectors, correlations — behaves as if nothing matched.
+- There is no re-scan or backfill API. The only way to evaluate an event against rules after the fact is [logtest](../content-manager/rule-testing.md), one event at a time.
+
+Plan a maintenance window that disables a detector accordingly, and re-ingest the source data if the interval matters.
+
+### Auditing state changes
+
+Every transition is recorded in the Wazuh Indexer log at `INFO`, together with the account that requested it:
+
+```
+[2026-09-10T13:54:15,906][INFO ][o.o.s.t.TransportIndexDetectorAction] [indexer] Detector [h1BSbaABJb1ilIZ5JvzP] (windows, standard) was disabled by [wazuh-admin]. No findings will be generated for its integration until it is enabled again.
+[2026-09-10T13:54:20,282][INFO ][o.o.s.t.TransportIndexDetectorAction] [indexer] Detector [h1BSbaABJb1ilIZ5JvzP] (windows, standard) was enabled by [wazuh-admin]. Events indexed while it was disabled are not re-evaluated.
+```
+
+A change carried with no authenticated user attached to the request — the Content Manager switching a detector off because its integration was disabled in the CTI catalog, or any cluster running without the security plugin — is logged the same way, with `internal` in place of the account name. Nothing is written when a request leaves the state unchanged, and nothing is written until the change is persisted.
+
 ## Wazuh enriched findings
 
 ### What is a finding?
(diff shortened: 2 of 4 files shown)


MauriFont (approved):
LGTM

Issue #1535, opened 2026-09-04
The two agent-reported indices map whatever an agent sends, and two agents already use 402 of the 1000-field budget


Rebits:
## Summary

`wazuh-agent-config` and `wazuh-agent-stats` are the only two Wazuh index templates shipped with `dynamic: true`; the other 42 are `strict` or `strict_allow_templates`. Both carry `total_fields.limit: 1000`. Their contents come straight from what an agent reports through `POST /config` and `POST /stats`, so every distinct key an agent sends becomes a permanent field in an index shared by the whole fleet. Two agents reporting their ordinary configuration already occupy **402 of the 1000** fields. Any authenticated agent can consume the rest, and once the limit is reached the config and stats reporting fails for every agent.

## The two exceptions

Scanning the 45 shipped index templates:

```
=== summary of dynamic
   Counter({'strict': 26, 'strict_allow_templates': 16, 'true': 2, None: 1})
```

and the two are:

```
=== states-agent-stats
   patterns: ['wazuh-agent-stats*']
   settings: {"mapping": {"total_fields": {"limit": "1000"}}, "refresh_interval": "5s", ...}
   dynamic: true   explicit fields: 3

=== states-agent-config
   patterns: ['wazuh-agent-config*']
   settings: {"mapping": {"total_fields": {"limit": "1000"}}, "refresh_interval": "5s", ...}
   dynamic: true   explicit fields: 3
```

Three explicit fields each, and everything else created on arrival.

## What two agents already cost

```
=== wazuh-agent-config
   dynamic=true  limit=1000  fields actuales=402
   fields with 4 or more levels: 394
   sample: ['message', 'state.document_version', 'state.modified_at',
             'wazuh.agent.configuration.content.agent-info.integrity_interval',
             'wazuh.agent.configuration.content.agent-info.interval',
             'wazuh.agent.configuration.content.agent-info.synchronization.enabled', ...]

=== wazuh-agent-stats
   dynamic=true  limit=1000  fields actuales=29
```

394 of the 402 are four levels deep or more, one per configuration option per module: `wazuh.agent.configuration.content.<module>.<option>`, and nested further for sub-blocks. The two agents in this deployment are the same platform with the same module set, so they contribute one copy of that tree between them. A fleet with Windows, macOS and Linux agents, or agents with different modules enabled, contributes one subtree per distinct shape.

## Any agent can add fields at will

[...]

PR #1575, opened 2026-09-14
Raise the default keyword ignore_above from 1024 to 4096

felipegonzalezmv:
## Description

A `keyword` value longer than the field's `ignore_above` is stored in `_source` but indexed nowhere. It matches no `term`, `wildcard`, `prefix` or even `exists` query, and no Sigma `|contains` selection can fire on it. The console still renders the full value out of `_source`, so nothing about the event looks wrong — the value is in plain sight and simultaneously invisible to detection.

WCS inherited ECS's default of `1024` on ~4,960 generated keyword fields, so padding any of them past 1 KB hides its contents from every rule that reads it.

Reported in wazuh/internal-devel-requests#6123 (QA, severity critical). Follows the plan agreed there: set `ignore_above` to `4096`.

The fields the report leads with — `message`, the six `*.command_line` variants, `url.original`/`full`/`path` — are already fixed: #1529 made them `match_only_text`, which has no length ceiling at all. What remained is the long tail of plain keyword fields: `user.name`, `process.args`, `process.name`, `file.path`, and ~1,600 more.

## Proposed Changes

The `1024` is not ours. It is ECS's own default, applied by a `setdefault` in `scripts/schema/cleaner.py` inside the WCS generator container. Changing it there and regenerating is one knob, rather than an `ignore_above` annotation on ~5,000 fields — and it keeps the generated docs (`wcs_flat.yml`, `fields.csv`) consistent with the templates.

| Obstacle | Fix |
|---|---|
| `1024` is injected by upstream ECS at generation time, not written in our tree | `wcs/generator/images/Dockerfile` patches `cleaner.py` after the ECS clone, behind an `ARG KEYWORD_IGNORE_ABOVE=4096` |
| A future `ECS_VERSION` bump could move that line and make the `sed` a silent no-op, regressing every template to `1024` | the `RUN` carries a `grep` guard, so the image build fails instead |
| 122 entries in `wcs/stateful/agent-config/fields/custom/wazuh.yml` declare `1024` explicitly, so a `setdefault` never reaches them | bumped in the field YAML to match |

Because the upstream call is a `setdefault`, every field that declares its own value keeps it. `url.query` stays at `2083`, the ECS `origin_url` family at `8192`, `code_signature.thumbprint_sha256` at `64`, `file.drive_letter` at `1`, and the deliberate `256` in agent-config is left alone. No field type changed anywhere in the diff.

### What this does not do

**This is mitigation, not elimination.** A value past 4,096 characters is still invisible, and for a `keyword` field there is no way around that — as stated in the issue. `4096` is the mid-ground: ECS has never raised its own default despite open requests ([elastic/ecs#1481](https://github.com/elastic/ecs/issues/1481), [elastic/ecs#105](https://github.com/elastic/ecs/issues/105)), and while `8192` is safe for Lucene, the cost of an 8× jump in indexed terms per keyword field is unmeasured.

**A `|contains` rule may still not fire after this merges.** #1529 was reopened for a second defect that lives in the security-analytics repo: `DetectorMonitorConfig.getRuleIndexMappingsByType()` registers query-index analysis overrides for `text` and `keyword` only, with no `match_only_text` → `rule_analyzer` overlay. Until that lands, `message|contains` fails for any value containing a space. Separate PR; noted here because it gates end-to-end verification of `|contains` behaviour.

### Results and Evidence

Verified on the 5.x Docker cluster (OpenSearch **3.6.0**, not a 4.x node) with the rebuilt `wazuh-indexer-setup` plugin installed. Probe script and raw output: `~/wazuh/scripts/evidence-6123/`.

`user.name` is the probe field because it is a plain ECS keyword that took the default. A data stream only adopts a new template on rollover, so `before` is backing index `000001` (written under the old template) and `after` is `000002` (created by a rollover under the new one) — same cluster, same probe.

| `user.name` value | `000001` — `ignore_above: 1024` | `000002` — `ignore_above: 4096` |
|---|---|---|
| 512 chars | indexed | indexed |
| 2,000 chars | **NOT INDEXED** — `exists`, `term`, `prefix` all return 0 | **indexed** |
| 5,000 chars | NOT INDEXED | NOT INDEXED — documented residual |

<details><summary>Raw probe output</summary>

```
--- ignore_above applied to keyword fields by the shipped index template ---
  keyword fields with ignore_above=1: 5
  keyword fields with ignore_above=2083: 4
  keyword fields with ignore_above=4096: 1629
  keyword fields with ignore_above=64: 8
  keyword fields with ignore_above=8192: 12
  user.name mapping: {"ignore_above": 4096, "type": "keyword"}

--- ignore_above on the LIVE backing index (what queries actually hit) ---
  .ds-wazuh-events-v5-system-activity-000001: user.name -> {"ignore_above": 1024, "type": "keyword"}
  .ds-wazuh-events-v5-system-activity-000002: user.name -> {"ignore_above": 4096, "type": "keyword"}

--- is the indexed value reachable by a query?  (before: index 000001) ---
  len=512   exists=1    term=1    prefix=1     -> indexed
  len=2000  exists=0    term=0    prefix=0     -> NOT INDEXED
  len=5000  exists=0    term=0    prefix=0     -> NOT INDEXED

--- is the indexed value reachable by a query?  (after: index 000002) ---
  len=512   exists=1    term=1    prefix=1     -> indexed
  len=2000  exists=1    term=1    prefix=1     -> indexed
  len=5000  exists=0    term=0    prefix=0     -> NOT INDEXED
```
</details>

**Regression probes — the diff changes nothing else.** Census over `plugins/setup/src/main/resources/templates/**`, before → after:

| `ignore_above` | before | after |
|---|---|---|
| `1024` | 4961 | **0** |
| `4096` | 0 | **4961** |
| `8192` | 28 | 28 |
| `2083` | 8 | 8 |
| `256` | 34 | 34 |
| `64` | 18 | 18 |
| `1` | 12 | 12 |

Diff audit over the whole regenerated tree — every non-context line is `1024` → `4096`:

```
$ git diff -U0 -- wcs plugins ':!*fields.csv' | grep '^[+-]' | grep -v '^[+-][+-]' | sort | uniq -c
   5070 + "ignore_above": 4096,        5070 - "ignore_above": 1024,
   3874 + ignore_above: 4096           3874 - ignore_above: 1024
```

- **No type changed anywhere**: grepping the diff for `"type"` returns zero hits.
- **#1529 preserved**: `events.json` still carries 28 `match_only_text` fields; `process.command_line` and `url.original` are unchanged.
- **No limit churn**: grepping the diff for `total_fields` / `nested_fields` / `max_docvalue_fields_search` returns zero hits. `ignore_above` does not affect the field count, so the 1000-field budget tracked in #1535 is unaffected.
- **`fields.csv` untouched**: `ignore_above` is not a CSV column.

**Build and test:**

```
./gradlew :wazuh-indexer-setup:build -x check            BUILD SUCCESSFUL
./gradlew :wazuh-indexer-setup:test                      56 tests,  0 failures, 0 errors
./gradlew :wazuh-indexer-content-manager:test           752 tests,  0 failures, 0 errors
./gradlew spotlessCheck                                  BUILD SUCCESSFUL (repo-wide)
```

### Artifacts Affected

- **`wazuh-indexer-setup`** — all index templates under `src/main/resources/templates/{streams,states,content}/` plus `cve.json`.
- **`wazuh-indexer-content-manager`** — the generated `cti-*-mappings.json` set and `internal-state-mapping.json`.
- **WCS generator image** — `wcs/generator/images/Dockerfile`.
- Packaging, default configuration and security configuration: **not touched**.

Scope is deliberately uniform: the default applies to every module — event and findings streams, the stateful `states/*` indices, the content-manager CTI indices, agent-config. Narrowing it to the detection-relevant streams would mean per-field YAML overrides and a much larger, more fragile diff.

### Configuration Changes

No new or changed runtime settings. One new build argument, `KEYWORD_IGNORE_ABOVE` (default `4096`), on the WCS generator image.

**Backwards compatibility:** index templates apply at index creation, so existing indices keep `1024` until they roll over. Data streams pick the new value up on their next rollover with no intervention; long-lived stateful indices (`wazuh-states-*`) need a reindex to benefit. Nothing breaks in the meantime — the change only widens what is indexed, so no existing query or document stops working.

### Documentation Updates

No endpoint changed, so `openapi.yml` and `docs/ref/modules/content-manager/api.md` do not apply.

Generated field documentation was regenerated with the templates: `wcs/*/docs/wcs_flat.yml`. `wcs/*/docs/fields.csv` is unchanged because `ignore_above` is not one of its columns.

### Tests Introduced

None. The change carries no Java logic — it is a generated-mapping value — so there is nothing a unit test could assert that the template diff does not already state. The existing suites (56 setup + 752 content-manager) were run as regression cover, since both plugins load these templates from their resources at startup, and both pass.

Verification is the cluster probe under `~/wazuh/scripts/evidence-6123/`, which reproduces the reporter's steps 3–5 and cleans up after itself.

## Review Checklist

- [x] Code changes reviewed
- [ ] No unresolved dependencies with other issues — depends on the security-analytics `match_only_text` overlay (#1529's reopened tail) for end-to-end `|contains` verification
- [x] Correct labels applied (`no-changelog`)


Jorgesnchz (approved):
LGTM

PR #1577, opened 2026-09-14
Set the ism_template timestamp when indexing ISM policies

### LINKED ISSUE #1533: A rolled-over write index was created outside its own policy, so everything written to it now is unmanaged
`wazuh-ai-assistant-sessions` is the only data stream on this deployment that has rolled over, and it has done so four times. The first four backing indices carry `ai-assistant-sessions-policy`. The fifth, which is the current write index, carries no policy at all, although the policy's own template pattern matches its name.

|---|---|---|---|---|
| 5.0.0-latest | `ai-assistant-sessions-policy` and the `wazuh-ai-assistant-sessions` data stream | Indexer | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

**Severity:** High. An unmanaged write index never rolls over again and is never deleted by retention, so the stream stops rotating at the point this happens and everything written from then on accumulates in one index forever. It is silent: the stream keeps accepting writes and the console shows nothing.

## Measured

```
   GET /_data_stream
   wazuh-ai-assistant-sessions   generacion=5   indices=5
                                 write=.ds-wazuh-ai-assistant-sessions-000005

   GET /_plugins/_ism/explain/*
   .ds-wazuh-ai-assistant-sessions-000001   policy=ai-assistant-sessions-policy  state=hot
   .ds-wazuh-ai-assistant-sessions-000002   policy=ai-assistant-sessions-policy  state=hot
   .ds-wazuh-ai-assistant-sessions-000003   policy=ai-assistant-sessions-policy  state=hot
   .ds-wazuh-ai-assistant-sessions-000004   policy=ai-assistant-sessions-policy  state=hot
   .ds-wazuh-ai-assistant-sessions-000005   policy=None                          state=None
```

The policy's template pattern is `['.ds-wazuh-ai-assistant-sessions-*', 'wazuh-ai-assistant-sessions*']`, so `000005` matches it. Four indices with the same name shape were given the policy and the fifth was not.

## It is not a timing artefact

```
   000001  created 2026-08-25T09:13:07Z   managed
   000002  created 2026-08-26T09:13:09Z   managed
   000003  created 2026-08-27T09:17:39Z   managed
   000004  created 2026-08-30T11:40:26Z   managed
   000005  created 2026-08-31T11:45:23Z   no policy
```

`000005` was 95 minutes old when measured. ISM assigns a template policy at index creation and its job interval is minutes, not hours, and `000004` a day earlier was assigned normally. So this is not a policy that has yet to be picked up.

## Why it stops the rotation

`000005` is the write index. A data stream rolls over when the policy on the *write* index reaches its rollover condition. With no policy on the write index there is no rollover condition, so:

- generation 6 is never created,
- `000005` grows without limit,
- the delete transition that removes AI session data after seven days never applies to it,
- and the four managed indices ahead of it will be deleted on schedule, leaving the unmanaged one as the only survivor.

## The other twenty-one streams cannot show this yet

Every other data stream is at generation 1, so none has rolled over and none can exhibit the problem. That is why it appears on the one stream that rotates fastest. On a production deployment the events and findings streams roll over constantly, and this is where the same gap would show first.

## Expected behavior

Every backing index a rollover creates receives the policy its template names, and an index whose name matches an `ism_template` pattern and carries no policy is reported somewhere an operator sees.

## Steps to reproduce

1. `GET /_data_stream`. `wazuh-ai-assistant-sessions` is at generation 5 with `.ds-wazuh-ai-assistant-sessions-000005` as its write index.
2. `GET /_plugins/_ism/explain/.ds-wazuh-ai-assistant-sessions-*`. The first four have `index.plugins.index_state_management.policy_id` set and the fifth does not.
3. `GET /_plugins/_ism/policies/ai-assistant-sessions-policy`. Its `ism_template.index_patterns` matches `.ds-wazuh-ai-assistant-sessions-*`.
4. `GET /_cat/indices/.ds-wazuh-ai-assistant-sessions-*?h=index,creation.date.string`. The unmanaged one is the newest.

## Related

[048](BUG_048-retention-never-runs-because-rollover-can-never-complete.md) is the other way rotation stops, with the policy present and the condition unreachable. [127](not-for-filing/duplicated/DUPLICATED_127-events-are-configured-to-be-deleted-an-hour-after-the-findings-that-point-at-them-are-kept-for-ninety-days.md) covers what the retention windows are. [181](BUG_181-the-eighteen-state-indices-have-no-retention-no-rollover-and-no-alias-so-they-grow-with-the-estate-forever.md) is the set of indices that never had a policy to lose.

ferooo23:
## Description

ISM applies an `ism_template` only to indices created after the template's `last_updated_time`, and our six policies never declared that field. When it is missing ISM fills it with the current time every time it parses the policy, so the check never succeeds and no index is attached automatically: every backing index a rollover creates is born unmanaged, its stream stops rotating and the retention never reaches it.

Closes #1533

## Proposed Changes

- `IndexStateManagement.indexPolicy()` copies the policy's `last_updated_time` into each of its `ism_template` entries before indexing it, which is what the ISM API does when a policy is created through it. Doing it in code rather than in the JSON files means a new policy cannot ship without it.
- The six policy files carried that timestamp in seconds (`1772122150`). ISM and our own guide expect epoch milliseconds, so they now read `1772122150000`.

### Results and Evidence

Validation: https://github.com/wazuh/wazuh-indexer-plugins/pull/1577#issuecomment-5667000836

### Artifacts Affected

The six ISM policy documents in `.opendistro-ism-config`. They are rewritten on every node startup, so an existing deployment picks the change up on the first restart after the upgrade.

### Configuration Changes

None.

### Documentation Updates

`docs/dev/plugins/setup.md`, the notes of "Creating a new ISM (Index State Management) policy".

### Tests Introduced

Three unit tests in `IndexStateManagementTests`.

## Review Checklist

- [x] Code changes reviewed


ferooo23:
## Validation 

Single node AIO, nightly packages, with the `setup` plugin built from this branch.

### Before

The template timestamp was not stored, so ISM stamped it with the current time on every read. Two consecutive calls, 52 seconds apart:

```
GET _plugins/_ism/policies/ai-assistant-sessions-policy?filter_path=policy.ism_template.last_updated_time
```
```json
{ "policy": { "ism_template": [ { "last_updated_time": 1789140680404 } ] } }
{ "policy": { "ism_template": [ { "last_updated_time": 1789140732690 } ] } }
```

And the index a rollover created was born unmanaged, with no policy at all:

```
GET _plugins/_ism/explain/.ds-wazuh-ai-assistant-sessions-*
```
```json
".ds-wazuh-ai-assistant-sessions-000008": {
  "index.plugins.index_state_management.policy_id": null,
  "enabled": null
}
```

It stayed that way until the next node restart, which is what adopted it, and the rollover that followed produced another orphan.

### After

The timestamp is stored and does not move between reads:

```
GET _plugins/_ism/policies/ai-assistant-sessions-policy?filter_path=policy.last_updated_time,policy.ism_template
```
```json
{
  "policy": {
    "last_updated_time": 1772122150000,
    "ism_template": [
      {
        "index_patterns": [".ds-wazuh-ai-assistant-sessions-*", "wazuh-ai-assistant-sessions*"],
        "priority": 0,
        "last_updated_time": 1772122150000
      }
    ]
  }
}
```

Rolling the stream over, with no restart in between:

```
POST wazuh-ai-assistant-sessions/_rollover
GET _plugins/_ism/explain/.ds-wazuh-ai-assistant-sessions-*
```
```json
".ds-wazuh-ai-assistant-sessions-000002": {
  "policy_id": "ai-assistant-sessions-policy",
  "enabled": true
},
"total_managed_indices": 2
```

Same on a second stream, with a different policy:

```
POST wazuh-metrics-normalization/_rollover
GET _plugins/_ism/explain/.ds-wazuh-metrics-normalization-*
```
```json
".ds-wazuh-metrics-normalization-000002": {
  "policy_id": "stream-metrics-policy",
  "enabled": true
},
"total_managed_indices": 2
```

And ISM says it did it through the template in both cases:

```
[17:54:55][o.o.i.i.ManagedIndexCoordinator] Index [.ds-wazuh-ai-assistant-sessions-000002] matched ISM policy template and will be managed by ai-assistant-sessions-policy
[17:56:56][o.o.i.i.ManagedIndexCoordinator] Index [.ds-wazuh-metrics-normalization-000002] matched ISM policy template and will be managed by stream-metrics-policy
```

`registerWithISM()` did not take part: its log lines are only from the node startup, none for either `000002`. That same startup also shows ISM matching the template for the first backing index of every stream, with no `Registered backing index` line for any of them, so ISM got there first and the manual registration found the job document already in place.

The policy is not only attached, it is running. A few minutes later the new index is evaluating its rollover condition, which is what the stream needs to keep rotating:

```json
".ds-wazuh-ai-assistant-sessions-000002": {
  "action": { "name": "rollover" },
  "step": { "name": "attempt_rollover", "step_status": "condition_not_met" },
  "info": {
    "message": "Pending rollover of index [index=.ds-wazuh-ai-assistant-sessions-000002]",
    "conditions": { "min_index_age": { "condition": "1d", "current": "16.1m" } }
  }
}
```

And nothing else was left behind. Every backing index of the 22 data streams carries its policy:

```
GET _plugins/_ism/explain/.ds-wazuh-*?filter_path=*.index,*.policy_id,total_managed_indices
```
```json
"total_managed_indices": 24
```

plugins/setup/src/test/java/com/wazuh/setup/index/IndexStateManagementTests.java
```diff
@@ -178,4 +180,77 @@ public void testFileIOException() throws IOException {
...
+
+        for (String name : shipped) {
+            Map<String, Object> policyFile =
+                    utils.fromFile(IndexStateManagement.POLICIES_PATH + name + ".json");
+            IndexStateManagement.applyIsmTemplateTimestamp(policyFile);
+
+            Map<String, Object> policy = (Map<String, Object>) policyFile.get("policy");
+            Object lastUpdatedTime = policy.get("last_updated_time");
+            assertNotNull(name + " declares no last_updated_time", lastUpdatedTime);
+
```
MauriFont, line 244:
```
assertTrue( 
    name + " declares a last_updated_time that is not in the past",
    ((Number) lastUpdatedTime).longValue() < System.currentTimeMillis()
);
```

MauriFont (changes requested):
## Verified

Tested the fix on a real deployment and the behavior is correct:

- `ism_template.last_updated_time` no longer drifts between two consecutive reads of
  the policy — it is persisted now, so the index-creation check has a stable value to
  compare against.
- Forcing a rollover on a stream produces a backing index that is attached to its
  policy on creation, with no node restart involved. That is the failure in #1533 gone.

The approach is right, too: doing it in `indexPolicy()` rather than in the JSON files
mirrors what the ISM API does on policy creation and keeps the invariant in one place.

## Changes requested

**The test asserts the timestamp is present, not that it is in the past.**
That is the invariant that actually broke here: the bug was never "the field is missing"
as such, it was "the field resolves to `now`", and a value ahead of the deployment clock
reproduces it exactly.

plugins/setup/src/test/java/com/wazuh/setup/index/IndexStateManagementTests.java
ferooo23: Applied: f173b4de148dab774633b2c141ee3d9b35729b97

MauriFont (approved):
LGTM

Issue #1534, opened 2026-09-04
The shipped retention keeps findings for ninety days, metrics for thirty, and the events they came from for one hour


Rebits:
Six index lifecycle policies ship with the product. They differ only in the age at which they delete. Findings are kept 90 days, metrics 30 days, active responses 3 days, and the events those findings were derived from are kept **one hour**. The raw event stream is kept **ten minutes**. The policies say so in their own descriptions.

|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-latest, OpenSearch 3.6.0 | Index State Management policies shipped with the indexer | Indexer | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 agent, single-node stack |

**Severity:** High. An alert without the event it came from cannot be investigated. The product keeps its own performance metrics 720 times longer than the security telemetry those metrics measure.

## The six policies

```
   stream-raw-events-policy         min_index_age: 10m  -> delete
   stream-events-policy             min_index_age: 1h   -> delete
   stream-active-responses-policy   min_index_age: 3d   -> delete
   ai-assistant-sessions-policy     min_index_age: 7d   -> delete
   stream-metrics-policy            min_index_age: 30d  -> delete
   stream-findings-policy           min_index_age: 90d  -> delete
```

Every one of them uses the identical rollover threshold:

```json
   "rollover": {"min_doc_count": 200000000, "min_primary_shard_size": "20gb", "copy_alias": false}
```

The descriptions are explicit. From `stream-events-policy`:

> ISM policy for events data streams. Rolls over backing indices based on size or document count, and **deletes indices older than 1 hour**.

From `stream-raw-events-policy`:

> ISM policy for raw events data stream. ... and **deletes indices older than 10 minutes**.

## Why this is the wrong way round

A finding is a claim: "malware detected in `/etc`", "failed authentication from 10.0.0.5", "a scheduled task was created". The event is the evidence: the log line, the file hashes, the command line, the account. Triage means opening the finding and reading the event behind it, which every finding supports by carrying `event.index` and `event.doc_id`, verified working in this collection.

Under this retention, a finding raised at noon can be opened for three months and the event it points at is gone by one o'clock. `event.doc_id` resolves for an hour and dangles for the remaining 89 days and 23 hours.

[...]

PR #1579, opened 2026-09-15
Document the real ISM retention contract for stream data streams

### LINKED ISSUE #1530: Retention never runs: five of the six ISM policies gate deletion behind a rollover that needs 200 million documents
**Issue title:** `[BUG] Retention never runs: five of the six ISM policies gate deletion behind a rollover that needs 200 million documents`

**Describe the bug**

Each stream data type ships an ISM policy whose `hot` state performs a `rollover` and then transitions to `delete` after a stated age. On five of the six, the rollover has **no age condition**: it fires only at 200,000,000 documents or a 20 GB primary shard.

ISM evaluates a state's transitions only once the state's actions complete. The rollover never completes, so the transition is never reached, so nothing is ever deleted. The retention window in each policy's own description never applies.

The sixth policy, for the assistant sessions, has `min_index_age: 1d` on its rollover and therefore works. That is what makes this a defect rather than a deliberate choice: the correct shape is present in the same file set.

**To Reproduce**

On any deployment small enough not to reach the rollover thresholds:

```bash
# what the policy promises
curl -sk -u admin:admin 'https://localhost:9200/_plugins/_ism/policies/stream-events-policy' \
  | python3 -c 'import sys,json; p=json.load(sys.stdin)["policy"]; print(p["description"])'

# what it is actually doing
curl -sk -u admin:admin \
  'https://localhost:9200/_plugins/_ism/explain/.ds-wazuh-events-v5-security-000001'
# state: hot, step: attempt_rollover, step_status: condition_not_met, rolled_over: false

# and how old the index is against the stated window
curl -sk -u admin:admin \
  '_cat/indices/.ds-wazuh-events-v5-security-000001?h=index,docs.count,store.size,creation.date.string'
```

**Expected behavior**

Rollover carries an age condition, so a backing index closes on a schedule regardless of volume and becomes eligible for deletion. `min_index_age` on the rollover should be shorter than the retention age it feeds, in the way `ai-assistant-sessions-policy` already does it.

The alternative, if unbounded retention is genuinely intended for these streams, is to remove the delete states and the descriptions that promise them, because at present the policies document behaviour the product does not have.

**Plugins**

**Host/Environment**

**Additional context**

**Observed**

The six shipped policies, with the rollover conditions and the deletion age each one promises:

| Policy | Rollover conditions | Deletes after |
| --- | --- | --- |
| `stream-events-policy` | 200M docs, 20gb | 1 hour |
| `stream-findings-policy` | 200M docs, 20gb | 90 days |
| `stream-metrics-policy` | 200M docs, 20gb | 30 days |
| `stream-raw-events-policy` | 200M docs, 20gb | 10 minutes |
| `stream-active-responses-policy` | 200M docs, 20gb | 3 days |
| `ai-assistant-sessions-policy` | 200M docs, 20gb, **1 day** | 7 days |

Only the last has an age condition on the rollover.

The policies say plainly what they intend, for example:

```
"ISM policy for events data streams. Rolls over backing indices based on size or
 document count, and deletes indices older than 1 hour."
```

Every backing index carrying data is pinned in `hot`, on the rollover step, with the condition unmet:

```
.ds-wazuh-events-v5-security-000001:   state=hot step=attempt_rollover (condition_not_met) rolled_over=False
.ds-wazuh-findings-v5-security-000001: state=hot step=attempt_rollover (condition_not_met) rolled_over=False
.ds-wazuh-metrics-agents-000001:       state=hot step=attempt_rollover (condition_not_met) rolled_over=False
```

And the events index is 23 times past the age at which its own policy says it should have been deleted:

```
created:  2026-08-25 09:13:03
age:      23.4 hours          (policy says: delete after 1 hour)
docs:     868
size:     288.3kb
```

288 kilobytes is not going to reach 20 GB, and 868 documents is not going to reach 200 million, so this index will exist until somebody removes it by hand.

### Why it matters more than "an index is bigger than intended"

**The stated windows are not conservative, they are inoperative.** An operator reading `stream-raw-events-policy` sees a 10-minute retention on raw events and reasonably concludes that raw event data does not persist. It persists indefinitely. The same applies to any compliance or privacy claim resting on the 1-hour events window.

**Disks fill quietly, over a long enough period.** Worth being precise about the size of this, because an earlier version of this file implied more than the measurements support.

This battery did fill its host disk to 95%, the indexer did set `read-only-allow-delete` on every index, and that is [040](BUG_040-findings-duplicated-after-an-indexer-write-outage.md)'s trigger. **That was not caused by this defect.** Measured on the running deployment:

```
total across all 23 ISM-managed stream indices:   3.7 MB
.wazuh-threatintel-vulnerabilities (CVE feed):  1511.5 MB   <- not ISM-managed
wazuh-threatintel-enrichments:                   127.0 MB   <- not ISM-managed
everything else:                                  ~8   MB
total Wazuh index data:                            1.6 GB
```

The data this defect would have deleted is 3.7 MB. The host filled from other things, and the two indices that dominate Wazuh's footprint are content, not streams, and are outside these policies entirely.

So the honest impact is a slow one. On a deployment with real event volume the streams are where growth lives, and nothing will ever remove them: the 1-hour events window and the 10-minute raw-events window are the two that would matter most and neither has ever fired. But this battery is not evidence of that, and the file should not have implied it was.

**Small deployments are the affected ones.** A large installation will cross 20 GB per shard and roll over, and its retention will work approximately as written. A demo, a lab, a small site or a single-node deployment never will. The behaviour is inverted from what an operator would expect: the installations least able to absorb unbounded growth are the ones where retention silently does nothing.

### Root cause

`rollover` is the only action in the `hot` state. ISM runs a state's actions in order and evaluates the state's `transitions` only after they complete. With no age condition, the action sits at `attempt_rollover` returning `condition_not_met` on every execution, so `hot` never finishes and `delete` is unreachable.

```json
"states": [
  {"name": "hot",
   "actions": [{"rollover": {"min_doc_count": 200000000,
                             "min_primary_shard_size": "20gb",
                             "copy_alias": false}}],
   "transitions": [{"state_name": "delete",
                    "conditions": {"min_index_age": "1h"}}]},
  {"name": "delete", "actions": [{"delete": {}}]}
]
```

Compare with the one that works, which adds a single key:

```json
"rollover": {"min_doc_count": 200000000, "min_index_age": "1d",
             "min_primary_shard_size": "20gb", "copy_alias": false}
```

### Verified in source, and why it survived validation

Read straight from the six shipped policy files in `wazuh-indexer-plugins` branch `5.0.0`, `plugins/setup/src/main/resources/policies/`:

```
ai-assistant-sessions-policy.json    rollover {"min_index_age": "1d", "min_doc_count": 200000000, "min_primary_shard_size": "20gb"}   hot -> delete {'min_index_age': '7d'}
stream-active-responses-policy.json  rollover {"min_doc_count": 200000000, "min_primary_shard_size": "20gb"}                          hot -> delete {'min_index_age': '3d'}
stream-events-policy.json            rollover {"min_doc_count": 200000000, "min_primary_shard_size": "20gb"}                          hot -> delete {'min_index_age': '1h'}
stream-findings-policy.json          rollover {"min_doc_count": 200000000, "min_primary_shard_size": "20gb"}                          hot -> delete {'min_index_age': '90d'}
stream-metrics-policy.json           rollover {"min_doc_count": 200000000, "min_primary_shard_size": "20gb"}                          hot -> delete {'min_index_age': '30d'}
stream-raw-events-policy.json        rollover {"min_doc_count": 200000000, "min_primary_shard_size": "20gb"}                          hot -> delete {'min_index_age': '10m'}
```

One of the six has an age condition on its rollover. The other five do not, and those five are the ones carrying product data.

Reproduce that listing with no deployment at all:

```bash
cd wazuh-indexer-plugins/plugins/setup/src/main/resources/policies
for f in *.json; do
  python3 - "$f" <<'EOF'
import json, sys
d = json.load(open(sys.argv[1])); p = d.get('policy', d)
for s in p.get('states', []):
    for a in s.get('actions', []):
        if 'rollover' in a: print(sys.argv[1], 'rollover', json.dumps(a['rollover']))
    for tr in s.get('transitions', []):
        print(sys.argv[1], s['name'], '->', tr.get('state_name'), tr.get('conditions'))
EOF
done
```

**Why validation did not catch it.** `wazuh-indexer-plugins#1279` ("ISM policies validation", closed 2026-06-15) set out to verify exactly this, with a plan whose third item reads "Old indices are deleted when conditions are met". All three boxes were ticked. The method is recorded in the closing comment:

> I modified the policy to speed up the transitions:
>  - `rollover.min_doc_count: 20`
>  - Transition conditions to the `delete` state: `min_index_age: 10m`.

Lowering `min_doc_count` to 20 makes the rollover action complete, which is the precise step that never completes in the shipped policy. With rollover completing, the transition is reached and deletion works, so the validation observed a policy that is not the one shipped. The defect is invisible to any test that speeds the policy up in that way, and visible to any test that waits on the shipped one.

That is worth stating in the fix, because the regression test has to run the shipped conditions, or assert on the policy documents rather than on behaviour.

**How to confirm on a running deployment**, which is the check that discriminates:

```bash
# every backing index with data is pinned in hot, on attempt_rollover, condition not met
curl -sk -u admin:admin 'https://localhost:9200/_plugins/_ism/explain/.ds-wazuh-events-v5-*?pretty' |
  grep -E '"state"|"name"|"action"|info|message' | head -40

# and the index is older than the retention its own policy promises
curl -sk -u admin:admin 'https://localhost:9200/_cat/indices/.ds-wazuh-*?v&h=index,docs.count,store.size,creation.date.string&s=creation.date'
```

On an affected deployment the `hot` state never advances and no `.ds-` index is ever deleted, regardless of how far past the stated retention it is.

### Tasks

- [ ] Add `min_index_age` to the rollover action of the events, findings, metrics, raw-events and active-responses policies, as `ai-assistant-sessions-policy` already has.
- [ ] Check the same policies as shipped by the non-container installers, since this was only observed on the container deployment.
- [ ] Reconcile each policy's description with what it does, whichever way the fix goes.
- [ ] Add a suite check that a backing index older than its policy's retention window no longer exists. On this deployment that check fails at 23 hours against a 1-hour window.

felipegonzalezmv:
## Description

Each stream data type ships an ISM policy whose `hot` state rolls over and then transitions to `delete` after a stated age. On five of the six the rollover has no age condition: it fires at 20 GB per primary shard or 200,000,000 documents. ISM evaluates a state's transitions only once the state's actions complete, so on a deployment that never reaches those thresholds the backing index stays in `hot` on `attempt_rollover` with `condition_not_met`, and nothing is ever deleted.

Triage on the issue concluded that the **mechanism is correct and the defect is documentation**. Size-driven rollover followed by deletion of the rolled index is a deliberate design: these five streams rotate by volume. `ai-assistant-sessions-policy` carries `min_index_age: 1d` on its rollover because 7-day conversation retention only means something with day-granularity rotation, not because the other five are missing a "correct shape".

What is genuinely wrong is that nothing — neither the policy descriptions nor the mdBook — states the contract the product actually has:

```
retention ≈ max(deletion age, time to fill 20 GB per primary shard)
```

The deletion age is a **floor**, not a ceiling. An operator reading `stream-raw-events-policy` sees a 10-minute retention and reasonably concludes raw event data does not persist. On a quiet deployment it persists indefinitely. The same applies to any compliance or privacy claim resting on the 1-hour events window.

This PR is documentation only. No policy, template or plugin behaviour changes.

Closes #1530

## Proposed Changes

| Obstacle | Fix |
|---|---|
| The six `description` fields say "deletes indices older than X", which reads as a retention ceiling | Reworded to name the rollover condition, the deletion age as a floor, and the indefinite-retention case |
| The mdBook documents retention nowhere — `ref/modules/setup/index.md` says only that stream indices "are rolled over periodically" | New page `docs/ref/modules/setup/retention.md` with the formula, per-stream table, disk sizing and the opt-in for a hard time ceiling |
| Nothing states that low-volume deployments retain events and raw events indefinitely | Explicit section, with the `_ism/explain` command that shows it, and a blockquote stating these ages cannot be used as privacy or compliance statements |
| `dev/plugins/setup.md` and seven WCS module READMEs repeat the same misleading phrasing | Corrected; `Retention Period` relabelled `Deletion age (floor)` and cross-linked to the new page |
| Editing a shipped policy — the procedure an operator would reach for — is silently reverted on restart | New section stating the mechanism and the partial-revert failure mode, and the supported alternative |

The new reference page also records the two decisions taken during triage, so they are not re-litigated later:

- **Do not add an age condition to the shipped rollover** (the report's ask). It would cap retention time: a quiet deployment would start losing events after a day where today it keeps months inside its 20 GB. Documented instead as an opt-in for deployments that need a regulatory or data-minimisation ceiling — as a **separate `policy_id` at `ism_template` priority 1**, never as an edit to a shipped policy, because the plugin rewrites its own six on every restart.
- **Do not use `min_rollover_age`.** It expresses "N after rollover", stacking the window on top of an already-full generation — no guarantee we do not already have, and more disk at exactly the ingest rates where disk matters.

What this does **not** do: it does not change the floor values (`1h` events, `10m` raw were confirmed on the issue), does not change rollover conditions, and does not touch shard or replica settings.

### Results and Evidence

**Before** — what the policy promised and what it was doing, on a deployment below the thresholds:

```console
$ curl -sk -u admin:admin 'https://localhost:9200/_plugins/_ism/policies/stream-events-policy' | jq -r .policy.description
ISM policy for events data streams. Rolls over backing indices based on size or document count, and deletes indices older than 1 hour.

$ curl -sk -u admin:admin 'https://localhost:9200/_plugins/_ism/explain/.ds-wazuh-events-v5-security-000001'
state: hot, step: attempt_rollover, step_status: condition_not_met, rolled_over: false

created:  2026-08-25 09:13:03
age:      23.4 hours          (policy says: delete after 1 hour)
docs:     868
size:     288.3kb
```

**After** — the same description now states the contract that produces that behaviour:

```console
$ jq -r .policy.description plugins/setup/src/main/resources/policies/stream-events-policy.json
ISM policy for events data streams. Rolls over backing indices at 20 GB per primary shard (or 200,000,000 documents), then deletes each rolled-over index once it reaches 1 hour of age. Effective retention is the greater of the two: the 1 hour is a floor, not a ceiling, and deployments that never fill a shard retain events indefinitely.
```

<details><summary>All six descriptions after the change</summary>

| Policy | Description |
|---|---|
| `stream-events-policy` | Rolls over at 20 GB per primary shard (or 200,000,000 documents), then deletes each rolled-over index once it reaches **1 hour** of age. Floor, not ceiling; never-filling deployments retain events indefinitely. |
| `stream-raw-events-policy` | Same rollover, deletion age **10 minutes**, same floor wording for raw events. |
| `stream-findings-policy` | Same rollover, deletion age **90 days**. |
| `stream-metrics-policy` | Same rollover, deletion age **30 days**. |
| `stream-active-responses-policy` | Same rollover, deletion age **3 days**. |
| `ai-assistant-sessions-policy` | Rolls over **daily**, or earlier at 20 GB per primary shard; deletes at **7 days**. The daily rollover bounds retention: the deletion age runs from index creation and an index only takes writes during its first day, so a session lives at most **7 days** regardless of volume. |

</details>

**No side effects — the policy documents are otherwise byte-identical.** Only the `description` value differs on each of the six files:

```console
$ git diff --stat 53031b94^ 53031b94
 .../policies/ai-assistant-sessions-policy.json     | 2 +-
 .../policies/stream-active-responses-policy.json   | 2 +-
 .../resources/policies/stream-events-policy.json   | 2 +-
 .../resources/policies/stream-findings-policy.json | 2 +-
 .../resources/policies/stream-metrics-policy.json  | 2 +-
 .../policies/stream-raw-events-policy.json         | 2 +-
 6 files changed, 6 insertions(+), 6 deletions(-)
```

Every state, action, transition, retry block, rollover condition and `ism_template` is untouched, and all six files still parse:

```console
$ for f in plugins/setup/src/main/resources/policies/*.json; do python3 -c "import json;json.load(open('$f'))" && echo "OK $f"; done
OK .../ai-assistant-sessions-policy.json
OK .../stream-active-responses-policy.json
OK .../stream-events-policy.json
OK .../stream-findings-policy.json
OK .../stream-metrics-policy.json
OK .../stream-raw-events-policy.json
```

**Why the documented opt-in changed (raised in review).** Reproduced on docker-env, OpenSearch 3.6.0, built from this branch; probe and raw runs in `~/wazuh/scripts/evidence-1530/`.

Editing a shipped policy is accepted and then silently reverted on the next restart, so the procedure this PR originally documented was not supported:

```console
### AFTER THE EDIT - what the operator sees
  description    : OPERATOR EDIT - 24h hard ceiling - do not overwrite
  rollover       : {'min_doc_count': 200000000, 'min_index_age': '1h', 'min_primary_shard_size': '20gb'}
  marker present : True

### AFTER A RESTART - what the operator gets back
  description    : ISM policy for events data streams. Rolls over backing indices at 20 GB per primary shard
  rollover       : {'min_doc_count': 200000000, 'min_primary_shard_size': '20gb'}
  marker present : False

[13:31:30] ISM policy [stream-events-policy] created      <- all six, on every boot
--- count of 'skipping creation' lines: 0
```

The replacement procedure was verified before being documented — a policy under its own `policy_id` with `ism_template` at `priority: 1` survives the restart, and a new backing index picks it up:

```console
### 5. Did the custom policy survive?
  present      : True
  rollover     : {'min_doc_count': 200000000, 'min_index_age': '1h', 'min_primary_shard_size': '20gb'}

### 6. Which policy is each events backing index managed by?
  .ds-wazuh-events-v5-security-000001 -> stream-events-policy
  .ds-wazuh-events-v5-security-000002 -> custom-events-24h-ceiling
```

That run also exposed the step the page now carries: `change_policy` is asynchronous, and restarting before the ISM sweep applies it drops the change, leaving indices split across two policies. Without a restart it lands:

```console
### AFTER change_policy + ISM sweep (no restart)
  .ds-...-000001  policy_id: custom-events-24h-ceiling
                  info: Attempting to change policy to custom-events-24h-ceiling
```

**The docs build and the new page is reachable:**

```console
$ cd docs && mdbook build
 INFO Book building has started
 INFO Running the html backend
 INFO HTML book written to `.../docs/book`

$ ls book/ref/modules/setup/
api-reference.html  architecture.html  configuration.html  index.html  retention.html  schema.html
```

Cross-links resolve in the rendered output — `retention.html` → `../../../dev/plugins/setup.html`, `index.html` → `retention.html`, and the four new links from the developer guide → `../../ref/modules/setup/retention.html`.

**Figures in the page are checked against the shipped templates**, not asserted: one primary shard and `auto_expand_replicas: 0-1` per stream template, so the ×2 replica column holds, and the 200M-doc condition never fires first (at ~1 KB/event that is 200 GB on a shard that rolls at 20 GB).

### Artifacts Affected

- **Plugin:** `wazuh-indexer-setup` — the six ISM policy resources under `src/main/resources/policies/`, `description` field only.
- **Packaging:** not affected.
- **Default configuration:** not affected.
- **Index templates:** not affected.
- **Security configuration:** not affected.

Note that the six shipped policies are **re-indexed on every cluster-manager election**, which includes every restart: `IndexStateManagement.initialize()` calls `createPolicies()` unconditionally and `indexPolicy()` writes each one by policy id with a plain `IndexRequest`, which upserts. An already-initialized cluster therefore picks up the new description text on its next restart, with no manual step — and, by the same mechanism, any operator edit to a shipped policy is reverted on that restart. That is why this PR documents creating a policy under a separate `policy_id` rather than editing a shipped one.

(An earlier revision of this section claimed the opposite — that policies are not overwritten. It was wrong on both halves; corrected after review.)

### Configuration Changes

None. No new settings, no changed defaults, no API change. Fully backwards compatible: the policy documents are semantically unchanged, so a cluster that applies the new files behaves exactly as before.

### Documentation Updates

- `docs/ref/modules/setup/retention.md` — **new**, the user-facing retention reference.
- `docs/SUMMARY.md` — registers the new page under Modules → Setup (an unlisted page is not built).
- `docs/ref/modules/setup/index.md` — links to it from the ISM bullet.
- `docs/dev/plugins/setup.md` — the four per-policy sections corrected, `Retention Period` → `Deletion age (floor)`, cross-linked.
- `wcs/ai-assistant/sessions/docs/README.md`, `wcs/stateless/active-responses/docs/README.md`, `wcs/stateless/events/findings/docs/README.md`, `wcs/stateless/events/raw/docs/README.md`, `wcs/stateless/metrics/agents/docs/README.md`, `wcs/stateless/metrics/comms/docs/README.md`, `wcs/stateless/metrics/engine/docs/README.md` — the ISM line in each corrected. `grep` over `wcs/` now returns no `ISM policy:` line without the rollover in it.
- `CHANGELOG.md` — entry under `## [v5.0.0]` → `### Fixed`.

No `openapi.yml` change: no endpoint is added or modified.

### Tests Introduced

None. The change is limited to a JSON `description` field and Markdown prose; there is no behaviour to assert. Existing setup plugin tests cover policy creation and are unaffected — no test asserts on policy description text.

Verification was done by build and inspection instead, as shown above: JSON validity on all six policies, a scoped diff proving no structural change, and an `mdbook build` with link resolution on the rendered HTML.

## Review Checklist

- [x] Code changes reviewed
- [x] Tests cover the new functionality — not applicable, documentation only
- [x] Configuration changes documented — none introduced
- [x] No unresolved dependencies with other issues — #1534 handled separately
- [x] Correct labels applied (e.g., `no-changelog`) — CHANGELOG entry added, label not needed


plugins/setup/src/main/resources/policies/ai-assistant-sessions-policy.json
```diff
@@ -1,7 +1,7 @@
 {
   "policy": {
     "policy_id": "ai-assistant-sessions-policy",
-    "description": "ISM policy for the AI assistant sessions data stream. Rolls over backing indices daily and deletes indices older than 7 days.",
+    "description": "ISM policy for the AI assistant sessions data stream. Rolls over backing indices daily, or earlier at 20 GB per primary shard (or 200,000,000 documents), then deletes each rolled-over index once it reaches 7 days of age. The daily rollover bounds retention, so sessions are kept for at most about 8 days regardless of volume.",
```
MauriFont: At most for 7 days, not 8.

MauriFont (changes requested):
Three things to fix.

### 1. The documented opt-in is not supported — it is reverted on every restart

`retention.md` (*Opting in to a hard time ceiling*) tells the operator to `PUT` an edited `stream-events-policy`. The Setup plugin re-indexes its six policy documents on every cluster-manager election: `IndexStateManagement.initialize()` calls `createPolicies()` unconditionally and `indexPolicy()` writes a plain `IndexRequest` with the policy id, so it upserts. The `catch (ResourceAlreadyExistsException)` that logs "Policy already exists, skipping creation" cannot fire for a document write.

Verified on 5.0.0: `PUT` with a marked description → HTTP 200 (`_version 22`); `systemctl restart wazuh-indexer` → `_version 23` and the shipped text is back. The log shows all six policies re-created on each boot, and never the "skipping creation" line.

It fails partially, which is worse: each managed index holds its own copy of the policy (`explain?show_policy=true`) and `StreamIndex.registerWithISM()` writes with `opType(CREATE)`, so indices already moved by `change_policy` keep the edit while every new backing index gets the shipped policy — two retention regimes, no signal.

What survives is a policy under its own id, so the section should document that instead. The steps are:

1. Create a new policy with its own `policy_id` — the plugin only ever rewrites its own six, so yours is never touched.
2. Put the `min_index_age` on the rollover action, shorter than the deletion age that follows it (same shape as `ai-assistant-sessions-policy`).
3. Give its `ism_template` the same index patterns with `priority: 1`, so it beats the shipped policies (`priority: 0`) for newly created backing indices.
4. Run `change_policy` to move the already-managed indices onto it.

### 2. The `Artifacts Affected` note states the opposite of what the plugin does

> Note that ISM policies are created on first initialization and are **not** overwritten on upgrade, so an already-initialized cluster keeps the old description text until the policy is updated by hand.

Both halves of that are wrong, for the reasons above.

### 3. Three WCS READMEs keep the misleading phrasing

Not in the diff, and the worst of the set — they do not mention the rollover at all:

- `wcs/stateless/metrics/agents/docs/README.md:17`
- `wcs/stateless/metrics/comms/docs/README.md:23`
- `wcs/stateless/metrics/engine/docs/README.md:17`

All three read `- **ISM policy:** `stream-metrics-policy` — deletes backing indices older than 30 days.` Suggested replacement, matching the other four:

> - **ISM policy:** `stream-metrics-policy` — rolls over at 20 GB per primary shard (or 200M docs) and deletes rolled-over indices once they reach 30 days of age. Retention is the greater of the two.

felipegonzalezmv:
@MauriFont thanks — the three points are all correct. I reproduced the first one on a 5.x cluster rather than taking it on reading, and the reproduction turned up one extra detail that affects the replacement procedure you proposed.

Rebased on `5.0.0` first, so the conflict with #1577 is gone: it touched the same six policy files, and the clash was line-for-line — my new `description` against its `last_updated_time` in milliseconds. Both sides kept.

---

### 1. Confirmed, and reproduced

Your reading of the code is right. `SetupPlugin` registers a `LocalNodeClusterManagerListener` whose `onClusterManager()` runs `indices.forEach(Index::initialize)`; `IndexStateManagement.initialize()` calls `createPolicies()` unconditionally, and `indexPolicy()` writes a plain `IndexRequest` with `.id(policy)`, which upserts.

Reproduced on docker-env (OpenSearch 3.6.0, cluster green, built from this branch). Probe and raw runs are in `~/wazuh/scripts/evidence-1530/`.

The operator edit is accepted:

```
### AFTER THE EDIT - what the operator sees
  description    : OPERATOR EDIT - 24h hard ceiling - do not overwrite
  rollover       : {'min_doc_count': 200000000, 'min_index_age': '1h', 'min_primary_shard_size': '20gb'}
  marker present : True
```

and is gone after `docker restart`:

```
### AFTER A RESTART - what the operator gets back
  description    : ISM policy for events data streams. Rolls over backing indices at 20 GB per primary shard
  rollover       : {'min_doc_count': 200000000, 'min_primary_shard_size': '20gb'}
  marker present : False
```

with all six re-created on boot and the `ResourceAlreadyExistsException` branch never taken, exactly as you said:

```
[13:31:30] ISM policy [stream-events-policy] created
[13:31:30] ISM policy [stream-findings-policy] created
... all six ...
--- count of 'skipping creation' lines: 0
```

The divergence you flag also shows up: the policy copy embedded in `.ds-wazuh-events-v5-security-000001` (`explain?show_policy=true`) still carried the *previous* description while the policy document already had the new one. The two copies do not track each other.

**Your replacement procedure works** — I verified it before documenting it. A policy under its own `policy_id` with `ism_template` at `priority: 1` survives the restart intact, and a newly created backing index picks it up:

```
### 5. Did the custom policy survive?
  present      : True
  rollover     : {'min_doc_count': 200000000, 'min_index_age': '1h', 'min_primary_shard_size': '20gb'}

### 6. Which policy is each events backing index managed by?
  .ds-wazuh-events-v5-security-000001 -> stream-events-policy
  .ds-wazuh-events-v5-security-000002 -> custom-events-24h-ceiling
```

**One addition to your step 4, found the hard way.** `change_policy` is asynchronous. In the run above, `000001` came back on the shipped policy because I restarted ~15 s after issuing it — the pending change was simply never applied. Without a restart it does land, on the next ISM sweep:

```
### BEFORE change_policy
  .ds-...-000001  policy_id: stream-events-policy  step: attempt_rollover condition_not_met
change_policy -> {'updated_indices': 2, 'failures': False}

### AFTER change_policy + ISM sweep (no restart)
  .ds-...-000001  policy_id: custom-events-24h-ceiling
                  info: Attempting to change policy to custom-events-24h-ceiling
```

So the documented steps need a fifth: wait for the sweep and confirm with `explain` that `policy_id` has actually flipped, before restarting anything. Otherwise an operator who follows the procedure and restarts straight away ends up in precisely the split state you describe — and would have no signal that it happened.

I will rewrite *Opting in to a hard time ceiling* around your four steps plus that one, and drop the `PUT`-the-shipped-policy version entirely.

### 2. Confirmed

The `Artifacts Affected` note states the opposite of the behaviour, and worse than you put it: policies are rewritten on **every cluster-manager election**, not only on upgrade. Rewriting it.

### 3. Confirmed

All three files are as you quote them. My sweep missed them because I grepped for `deletes indices older than` and these read `deletes backing indices older than`. Taking your suggested replacement verbatim. With those three the set is complete — no `ISM policy:` line under `wcs/` keeps the old phrasing.

---

One thing outside this PR's scope, for a separate issue if you agree: the `catch (ResourceAlreadyExistsException)` in `indexPolicy()` is unreachable for a document write, and its log line ("Policy already exists, skipping creation") describes behaviour the method does not have. It reads as an intent to create-if-absent that was never implemented — which is arguably the root of all of point 1. Happy to open it.

plugins/setup/src/main/resources/policies/ai-assistant-sessions-policy.json
felipegonzalezmv: Right, 7. I had added the rollover age to the deletion age, which is not how it composes: the transition's `min_index_age` runs from **index creation**, not from rollover, and the index only takes writes during its first day. So the oldest session in an index is 7 days old when it is deleted and the newest is 6 — the bound is 7 days flat, and the daily rollover is what keeps the *newest* data from outliving it, not something that adds to it.

Fixed in 6663864, with the reasoning in the text so the next reader does not redo the same wrong sum:

> The daily rollover bounds retention: the deletion age runs from index creation and an index only takes writes during its first day, so a session is kept for at most 7 days regardless of volume.

Also corrected the same claim in the PR description.

felipegonzalezmv:
@MauriFont all four points are addressed in 6663864, one commit, docs only — no code change, as agreed. Summary of what landed:

**1. The unsupported opt-in is gone.** The `PUT`-the-shipped-policy procedure is replaced by your four steps, plus one I had to add after reproducing it (below). The section now opens with a new one, *The shipped policies are rewritten on every restart*, stating the mechanism and the partial-revert failure mode, and ending with a flat "never edit one of the six shipped policies in place".

The extra step is **wait for `change_policy` to actually apply, and verify it**. It is asynchronous: the response reports the indices it queued, not indices it changed, and the switch lands on the managed index's next ISM run about a minute later. In my first run I restarted ~15 s after issuing it and the pending change was simply dropped — `000001` came back on the shipped policy while `000002` was on mine. That is your split state, produced by following the procedure. Without the restart it applies cleanly:

```
### AFTER change_policy + ISM sweep (no restart)
  .ds-...-000001  policy_id: custom-events-24h-ceiling
                  info: Attempting to change policy to custom-events-24h-ceiling
```

So the page now says explicitly: every index must report your `policy_id` in `explain` before you restart anything.

**2. The `Artifacts Affected` note is rewritten** in the PR description, and says what actually happens — re-indexed on every cluster-manager election, so an initialized cluster picks up the new text on its next restart with no manual step, and by the same mechanism an operator edit is reverted. I left a line marking that the earlier claim was wrong, rather than quietly swapping it.

**3. The three metrics READMEs** take your suggested replacement verbatim. `grep` over `wcs/` now returns no `ISM policy:` line without the rollover in it.

**4. The 7-day bound**, answered inline.

The developer guide also gained a short note on the overwrite next to the `last_updated_time` guidance, since that is where someone adding a policy will be reading.

One thing deliberately **not** in this PR, since we agreed code stays out: the `catch (ResourceAlreadyExistsException)` in `indexPolicy()` is unreachable for a document write, and its log line claims a create-if-absent behaviour the method does not have. Tell me if you want the issue and I will open it.

MauriFont (approved):
LGTM

PR #1580, opened 2026-09-16
Retry shed bulk updates instead of aborting the CTI consumer sync

### LINKED ISSUE #1900: The CVE consumer aborts on a parent circuit-breaker rejection, and the `iocs` consumer is rate-limited by the CTI API
## Description

Detected during [2026-09-10 Nightly](https://github.com/wazuh/devel-major-incidents/issues/248). 

The Central Components install suite fails `test_00_setup_install.py::test_check_logs[Indexer]` on
two message classes emitted during the initial CTI catalog synchronization. **Neither is in the
known-logs catalog (wazuh/wazuh-qa-automation#8672) nor tracked by any open issue**, so both are
reported here for the indexer team to classify: finding 1 looks like a genuine defect, and finding 2
we cannot classify ourselves — we need a verdict on whether it is expected.

Both come from the same run: nightly of 2026-09-10, build [1129](https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/1129/) (DIT-CC, 10 clustered AIO nodes), `5 failed, 88 passed`. Evidence is from `full_component_logs.zip`, `aio-amazon-2023-arm64_indexer.log` (`node-4`, the cluster-manager node
that performs the sync).

---

## Finding 1 — `UpdateServiceImpl` aborts the whole CVE consumer sync on a parent circuit-breaker rejection

A bulk **update** rejected by the parent circuit breaker propagates all the way up and abandons the
entire `cti:catalog:consumer:vulnerabilities` synchronization with three `ERROR` lines.

The asymmetry is the point: **the plugin already handles this exact condition gracefully on the
index path.** `ContentIndex` logs `Bulk indexing shed N document(s) under load; retry i/3 in Xms` at
`WARN` and retries — the accepted behaviour introduced in #1488 and analysed in
wazuh/wazuh-indexer#1851. `UpdateServiceImpl` has no equivalent: one `CircuitBreakingException` and
the sync is abandoned. Both happened on the same node **48 seconds apart**, same root condition (JVM
heap at its limit), handled two different ways in two code paths.

The index path sheds load and retries, at `WARN`:

```
[2026-09-10T04:00:20,443][WARN ][c.w.c.c.c.i.ContentIndex ] [node-4] Bulk indexing shed 797 document(s) under load; retry 1/3 in 1000ms.
[2026-09-10T04:00:20,495][WARN ][c.w.c.c.c.i.ContentIndex ] [node-4] Bulk indexing shed 140 document(s) under load; retry 1/3 in 1000ms.
```

48 seconds later the update path hits the parent breaker and gives up, at `ERROR`:

```
[2026-09-10T04:00:50,266][INFO ][c.w.c.c.c.s.AbstractConsumerService] [node-4] Updating consumer [cti:catalog:consumer:vulnerabilities] content (offset 906452 → 976063).
[2026-09-10T04:01:08,252][ERROR][c.w.c.c.c.s.UpdateServiceImpl] [node-4] Failed to apply offset [912496] (type=UPDATE, resource=CVE-2024-35255): Bulk update failed for document [CVE-2024-21810]: RemoteTransportException[[node-8][172.31.45.118:9300][indices:data/write/bulk[s]]]; nested: CircuitBreakingException[[parent] Data too large, data for [indices:data/write/bulk[s]] would be [1644383646/1.5gb], which is larger than the limit of [1639134003/1.5gb], real usage: [1641840928/1.5gb], new bytes reserved: [2542718/2.4mb], usages [request=0/0b, fielddata=0/0b, in_flight_requests=2542718/2.4mb]];
[2026-09-10T04:01:08,258][ERROR][c.w.c.c.c.s.UpdateServiceImpl] [node-4] Error during content update: Bulk update failed for document [CVE-2024-21810]: RemoteTransportException[[node-8][172.31.45.118:9300][indices:data/write/bulk[s]]]; nested: CircuitBreakingException[[parent] Data too large, ...];
[2026-09-10T04:01:08,326][ERROR][c.w.c.j.j.CatalogSyncJob ] [node-4] Error during synchronization of ConsumerCveService: Synchronization failed for consumer [cti:catalog:consumer:vulnerabilities]
[2026-09-10T04:01:08,328][WARN ][c.w.c.j.j.CatalogSyncJob ] [node-4] Synchronization failed; triggering one immediate retry.
```

The heap pressure is real and confirmed by the breaker service itself, nine seconds later:

```
[2026-09-10T04:01:17,327][INFO ][o.o.i.b.HierarchyCircuitBreakerService] [node-4] attempting to trigger G1GC due to high heap usage [1747189776]
[2026-09-10T04:01:17,459][INFO ][o.o.i.b.HierarchyCircuitBreakerService] [node-4] GC did bring memory usage down, before [1747189776], after [972813512], allocations [272], duration [132]
```

The built-in immediate retry then **does** complete, five minutes later:

```
[2026-09-10T04:01:08,915][INFO ][c.w.c.c.c.s.AbstractConsumerService] [node-4] Updating consumer [cti:catalog:consumer:vulnerabilities] content (offset 912446 → 976063).
[2026-09-10T04:06:04,679][INFO ][c.w.c.c.c.s.UpdateServiceImpl] [node-4] Successfully updated consumer [cti:catalog:consumer:vulnerabilities] to offset [976063]
```

So no content is lost. What the abort costs is three `ERROR` lines and about five minutes, on a path
that already knows how to back off.

### The condition is not isolated to one node

Counted across the ten `*_indexer.log` files of the same run:

| Node | `CircuitBreakingException` | `attempting to trigger G1GC` |
| :-- | :-: | :-: |
| `amazon-2023-arm64` (node-4) | 3 | 5 |
| `ubuntu-24-amd64` | 0 | 5 |
| `ubuntu-24-arm64` | 0 | 11 |

Three of ten nodes ran close enough to the heap limit to trigger G1GC during the setup window; the
breaker tripped on the one that was doing the CVE update at the time. A node with less headroom, or
a larger offset delta, can plausibly fail the retry too.

### Not a duplicate of #965

**#965 — "Too large CVE feed documents are rejected"** (open) is the same family (oversized CVE
content overwhelming the indexer) but a **different mechanism**: #965 is an indexing-pressure
rejection (`rejected execution of coordinating operation [coordinating_and_primary_bytes=…,
max_coordinating_and_primary_bytes=…]`) whose consequence is `ResourceNotFoundException: Document
with ID '…' could not be found in any ContentIndex`. Here the document is found, and the rejection
comes from the **parent circuit breaker** (JVM heap) on the bulk update. If you consider the two one
problem, this can be folded into #965 — but the proposed change below is specific to the update
path's error handling.

### Proposed changes

1. **Treat a `CircuitBreakingException` on the bulk-update path as retryable**, the same way
   `ContentIndex` treats a shed bulk index: back off and retry at `WARN`, and only escalate to
   `ERROR` once the retries are exhausted. Today the first rejection aborts the consumer sync.
2. **Bound the bulk-update batch by estimated payload size**, so the parent breaker is not the
   mechanism that limits the batch.
3. If the current abort-then-retry behaviour is intended as it stands, say so and we will catalogue
   the three lines under wazuh/wazuh-qa-automation#8672. Note the level is arguable either way: an
   error that self-heals on the immediate retry reads as a `WARN`.

---

## Finding 2 — the `iocs` consumer is rate-limited by the CTI API: is this expected?

**We cannot classify this one and are not trying to.** The message is not in the known-logs catalog
and no open issue tracks it, so it needs a verdict from the team: either it is expected during a
normal setup and we catalogue it, or the rate-limiting itself is a problem to fix.

```
[2026-09-10T03:56:17,475][WARN ][c.w.c.c.c.c.ApiClient    ] [node-4] CTI API rate-limited request [/api/v1/catalog/contexts/t1-iocs-5/consumers/public-iocs-5/changes?from_offset=353074&to_offset=354073] (HTTP 429); retrying in 60s (attempt 1/3).
[2026-09-10T03:57:17,482][WARN ][c.w.c.c.c.c.ApiClient    ] [node-4] CTI API rate-limited request [/api/v1/catalog/contexts/t1-iocs-5/consumers/public-iocs-5/changes?from_offset=353074&to_offset=354073] (HTTP 429); retrying in 60s (attempt 2/3).
```

What the log establishes:

- Both 429s are on the **same request** (same `from_offset=353074&to_offset=354073`), logged as
  `attempt 1/3` and `attempt 2/3`, sixty seconds apart. The counter and the delay are both correct.
- The consumer **recovered**: `[03:58:36,878] Successfully updated consumer
  [cti:catalog:consumer:iocs] to offset [365017]`. The third retry was never needed.
- **Only one node is affected**, and only one node does the work: 2 occurrences on `node-4`, 0 on the
  other nine. So this is not ten nodes competing against the API — it is a single client doing
  sequential chunked fetches and being throttled twice on one chunk.
- The endpoint is the pre-production CTI API (`api.pre.cloud.wazuh.com`, per the `SnapshotClient`
  lines in the same log).

Questions we need answered to classify it:

1. Is a 429 on a routine incremental `changes` fetch an expected, benign condition for a single
   syncing node, or does it indicate a problem on the client or the API side?
2. Should the retry emit `WARN` when it goes on to succeed, or is `INFO` the right level for a
   throttle that is absorbed transparently?
3. The `sync_interval` question from #1830 is still open in practice: had the third retry also been
   throttled, the consumer would have been left `status: failed` until the next scheduled sync. Is
   three retries at a fixed 60 s the intended budget for this endpoint?

### Relationship to the previous 429 reports — and why this is not a recurrence

This is the same *symptom* as wazuh/wazuh-indexer#1807, #1819 and #1830, all closed. It is **not** a
recurrence of #1830 specifically: that issue's defect was retry delays of **60 s, then 0 s, then
0 s**, burning all attempts in under 90 seconds and leaving the consumer at `status: failed` until
the next scheduled sync (up to 60 minutes). Here both retries waited the full 60 s and the consumer
completed, so that fix is behaving as intended.

What is left over is the question above: the throttling still happens, it is still logged at `WARN`,
and nothing catalogues it — so it fails the CC suite every night regardless of whether it is benign.

---

Logs:
- [full_component_logs.zip](https://github.com/user-attachments/files/32052849/full_component_logs.zip)
- [logs.zip](https://github.com/user-attachments/files/32052864/logs.zip)

### LINKED ISSUE #1913: A single bulk failure during a rolling restart is treated as non-retryable and aborts the whole `iocs` snapshot
## Description

In the Central Components deployability suite, the node that owns the `cti:catalog:consumer:iocs` sync downloads the remote snapshot and indexes it into `wazuh-threatintel-enrichments-a`. While the suite performs its rolling indexer restart, three bulk batches of 999 documents are classified as **non-retryable** and dropped. The consumer then abandons the entire snapshot — 2997 documents — does not advance `local_offset`, and rolls back to the stable snapshot shipped with the plugin.

No further attempt is made for the rest of the run: there is no `Successfully updated consumer [cti:catalog:consumer:iocs]` line in **any** of the ten indexer logs, so the deployment finishes with the IOC content at whatever the bundled stable snapshot contains.

Logs:
- [full_component_logs.zip](https://github.com/user-attachments/files/32229944/full_component_logs.zip)
- [logs.zip](https://github.com/user-attachments/files/32229966/logs.zip)

## Environment

- Build: [#1162](https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/1162/) (`B1162_DIT-CC-5.0.0-rc1-internal-latest`), 2026-09-15
- Cluster: 10 indexer nodes, one per OS/arch (amazon-2023, redhat-9, redhat-10, ubuntu-24, ubuntu-26, x amd64/arm64)
- Affected node: `node-9` (`aio-ubuntu-26-arm64`), the node running the iocs sync
- Evidence file: `aio-ubuntu-26-arm64_indexer.log` inside `full_component_logs.zip` of that build
- Trigger context: `test_02_restart.py` restarts the indexer node by node, roughly every 30 s, between 04:53:11 and 04:58:31

## Error

The three dropped batches (`aio-ubuntu-26-arm64_indexer.log:3537-3540`), all identical except for the millisecond:

```
[2026-09-15T04:56:51,827][ERROR][c.w.c.c.c.i.ContentIndex ] [node-9] Bulk indexing dropped 999 document(s) with non-retryable failures. Last failure: [wazuh-threatintel-enrichments-a/FyGl4YIIT3CFRKZzAP2rIg] IndexNotFoundException[no such index [wazuh-threatintel-enrichments-a]]
```

The consequence, 25 s later (`aio-ubuntu-26-arm64_indexer.log:3673-3674`):

```
[2026-09-15T04:57:16,516][ERROR][c.w.c.c.c.s.SnapshotServiceImpl] [node-9] Snapshot indexing for consumer [cti:catalog:consumer:iocs] dropped 2997 document(s); local_offset was not advanced so the next sync retries the snapshot.
[2026-09-15T04:57:16,519][WARN ][c.w.c.c.c.s.AbstractConsumerService] [node-9] Remote snapshot failed for consumer [cti:catalog:consumer:iocs]. Rolling back from stable snapshot [/usr/share/wazuh-indexer/plugins/wazuh-indexer-content-manager/snapshots/iocs.stable.zip].
```

3 x 999 = 2997, so those three batches are the whole loss.

## Analysis

The failure is transient and cluster-topology-related, yet it is classified as non-retryable:

- The bulk names the index by name **and UUID**: `wazuh-threatintel-enrichments-a/FyGl4YIIT3CFRKZzAP2rIg`. That UUID was superseded on the same node at 04:46:24,957 by `eVOmo9zNRMWyH2UebJ9apA`, ten minutes before the failure, and `eVOmo9zNRMWyH2UebJ9apA` was itself deleted at 04:57:16,554 (`MetadataDeleteIndexService ... deleting index`) and recreated as `GpsAup28S-akgQUic8exqQ` at 04:57:17,478. So the write was resolved against an index generation that no longer existed.
- At the same instant, `node-8` left the cluster as part of the rolling restart: `[2026-09-15T04:56:51,998][INFO ][o.o.c.s.ClusterApplierService] [node-9] removed {{node-8}...}`, followed at 04:56:52,609 by `[wazuh-threatintel-enrichments-a][0] marking unavailable shards as stale`. The index is created with 1 shard and its replica count is only raised afterwards, so losing that node leaves the shard unavailable.

Which of the two is the trigger is **not established** by the log: an `IndexNotFoundException` carrying a concrete UUID points at the stale-generation reading, while the 170 ms gap to the node departure points at the availability reading. Either way the condition is transient and the snapshot could have been retried rather than discarded.

The 29 further `Bulk indexing dropped ... NodeClosedException[node closed {node-9}...]` lines between 04:57:23,284 and 04:57:24,013 are a different, well-understood event: node-9 shutting down for its own restart. They are already tracked as expected restart noise and are not what this report is about.

## What is not the cause

The manager is down for the whole run in this build (it fails to start without `etc/certs/remoted.pem`, https://github.com/wazuh/wazuh-qa-automation/issues/8874), but it is not involved here: the manager restarts only begin at 04:58:43, one to five minutes after these messages, and the messages appear on `node-9` alone while the manager is down on all ten nodes.

## Frequency

New on 2026-09-15. `SnapshotServiceImpl`, `AbstractConsumerService` and `CatalogSyncJob` appear zero times in the aggregated `testing-*.log` of the three previous Central Components install builds: [#1151](https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/1151/) (2026-09-12), [#1140](https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/1140/) (2026-09-11) and [#1129](https://jenkins-production.qa.wazuh.info/job/5.0.0/job/deployability_testing/1129/) (2026-09-10). Builds before 2026-09-09 no longer have retrievable artifacts.

## Secondary observation

Two nodes download the same snapshot within two seconds of each other, and a manual trigger collides with a run already in progress:

```
[2026-09-15T04:46:23,282][INFO ][c.w.c.c.c.c.SnapshotClient] [node-9] Starting snapshot download from [https://api.pre.cloud.wazuh.com/store/contexts/t1-iocs-5/consumers/public-iocs-5/374321_1789053807.zip]
[2026-09-15T04:46:25,279][INFO ][c.w.c.c.c.c.SnapshotClient] [node-1] Starting snapshot download from [https://api.pre.cloud.wazuh.com/store/contexts/t1-iocs-5/consumers/public-iocs-5/374321_1789053807.zip]
[2026-09-15T04:53:32,204][WARN ][c.w.c.j.j.CatalogSyncJob ] [node-9] Attempted to trigger CatalogSyncJob manually while it is already running.
```

The same duplication happens for the `ruleset` consumer at 04:45:49. Whether one download per node is by design is worth confirming; if it is not, the collision warning and the duplicated download are the same defect.

## Steps to reproduce

1. Deploy the clustered Central Components stack and let the content manager start its `cti:catalog:consumer:iocs` snapshot sync.
2. Restart the indexer nodes one by one while the snapshot is being indexed.
3. Observe the bulk batches failing with `IndexNotFoundException`, the whole snapshot being dropped and rolled back, and no later sync attempt within the run.

## Expected behavior

A bulk failure caused by an index being recreated, or by the node holding its shard leaving the cluster, is treated as retryable: the snapshot is retried rather than discarded, and the consumer reaches its remote offset instead of falling back to the bundled stable snapshot.

## Related

- https://github.com/wazuh/wazuh-indexer-plugins/pull/1488 and https://github.com/wazuh/wazuh-indexer/issues/1851 (closed; the accepted `Bulk indexing shed ...` behaviour under heap pressure, a different condition)
- The ruleset-consumer detector failure seen in the same run is reported separately; it is a different consumer and a different phase.

Jorgesnchz:
## Description

During Nightly testing it was detected that, one bulk item rejected by the parent circuit breaker abandoned the entire `cti:catalog:consumer:vulnerabilities` synchronization, with three `ERROR` lines:

``` console
[2026-09-10T04:01:08,252][ERROR][c.w.c.c.c.s.UpdateServiceImpl] [node-4] Failed to apply offset [912496] (type=UPDATE, resource=CVE-2024-35255): Bulk update failed for document [CVE-2024-21810]: RemoteTransportException[[node-8][172.31.45.118:9300][indices:data/write/bulk[s]]]; nested: CircuitBreakingException[[parent] Data too large, data for [indices:data/write/bulk[s]] would be [1644383646/1.5gb], which is larger than the limit of [1639134003/1.5gb], real usage: [1641840928/1.5gb], new bytes reserved: [2542718/2.4mb], usages [request=0/0b, fielddata=0/0b, in_flight_requests=2542718/2.4mb]];
[2026-09-10T04:01:08,258][ERROR][c.w.c.c.c.s.UpdateServiceImpl] [node-4] Error during content update: Bulk update failed for document [CVE-2024-21810]: RemoteTransportException[[node-8][172.31.45.118:9300][indices:data/write/bulk[s]]]; nested: CircuitBreakingException[[parent] Data too large, ...];
[2026-09-10T04:01:08,326][ERROR][c.w.c.j.j.CatalogSyncJob ] [node-4] Error during synchronization of ConsumerCveService: Synchronization failed for consumer [cti:catalog:consumer:vulnerabilities]

```

48 seconds earlier, on the same node, the *index* path met the same condition and handled it correctly, at `WARN`, with a retry:

``` console
[2026-09-10T04:00:20,443][WARN ][c.w.c.c.c.i.ContentIndex ] [node-4] Bulk indexing shed 797 document(s) under load; retry 1/3 in 1000ms.
```

That asymmetry is the bug. `ContentIndex` held three write paths that never agreed on a failure
policy:

| Path | Used by | Retry policy before this PR |
|------|---------|------------------------------|
| `submitBulk` / `retryOrDrop` (async) | snapshot load | Retries, but only recognises load shedding; a cluster-topology change is classified permanent and dropped |
| `update(id, ops, offset)` (single doc) | `applyOffset` UPDATE | Retries on `isCircuitBreakerException` only — narrower than `isRetryable`, misses rejected-execution and 429/503 |
| `executeBulkUpdate` (batched) | `UpdateServiceImpl` via `batchUpdate` | **None** — the first failed item throws and aborts the consumer sync |

Another error that has been detected is that during a rolling restart the index a batch had been resolved against was recreated, the `IndexNotFoundException` was classified permanent, and the `iocs` consumer discarded. Both issues are one policy disagreement, so both are fixed here together.

``` console
[2026-09-15T04:56:51,827][ERROR][c.w.c.c.c.i.ContentIndex ] [node-9] Bulk indexing dropped 999 document(s) with non-retryable failures. Last failure: [wazuh-threatintel-enrichments-a/FyGl4YIIT3CFRKZzAP2rIg] IndexNotFoundException[no such index [wazuh-threatintel-enrichments-a]]
[2026-09-15T04:57:16,516][ERROR][c.w.c.c.c.s.SnapshotServiceImpl] [node-9] Snapshot indexing for consumer [cti:catalog:consumer:iocs] dropped 2997 document(s); local_offset was not advanced so the next sync retries the snapshot.
```

In order to fix this problem this PR:
- Gives every write one classifier and two budgets: 3 retries over 7 s(1s-2s-4s) for a cluster shedding load, which clears in seconds, and 5 retries over 95 s (5s-10s-20s-30s-30s) for a cluster-topology change, which has to outlast a node restart.
- Retries the items a bulk update sheds, logging each retry at `WARN`; throws only once the budget is exhausted.
- Retries a transient topology failure on the snapshot path rather than discarding the whole snapshot.
- Keeps aborting immediately on a permanent failure, leaving the consumer checkpointed at the last offset that actually landed. Nothing is ever dropped.
- Retries a whole-request rejection (`ExecutionException`) and the `MultiGet` that opens `batchUpdate` — both aborted the sync just as readily, with no retry at all.
- Collapses `update()`'s narrower `isCircuitBreakerException` check into the shared classifier, so every synchronous write uses one definition of "this is worth trying again".

Closes https://github.com/wazuh/wazuh-indexer/issues/1900
Closes https://github.com/wazuh/wazuh-indexer/issues/1913

> [!NOTE]
>  The values for the schedules dome from:
> 
> Shed: 3 retries, 1 s → 2 s → 4 s (7 s)
> Unchanged: these are the constants the indexer was already using
>
> Topology: 5 retries, 5 s → 10 s → 20 s → 30 s → 30 s (95 s)
> Derived from the issue report, the 30 s max comes from `BULK_MAX_BACKOFF_MS`, this is why we don't have   5 s → 10 s → 20 s → 40 s → 80 s

## Proposed Changes

**`ContentIndex.java`**

- A write the cluster refuses is no longer abandoned on the first refusal. The failed writes are re-submitted a few times, with a growing pause between attempts, and the operation fails only once those attempts are used up.
- How long it keeps trying depends on what went wrong: seconds for a cluster shedding load, long enough to outlast a node restart for a topology change. A batch holding both is retried on the schedule that outlasts both.
- Only the writes that were actually refused are re-sent, not the whole batch.
- The same treatment covers the read that opens an update, and the case where an entire request is turned away before any individual write is looked at. Both used to abandon the synchronization just as readily.
- A refusal that is genuinely permanent still fails immediately, and a node that is shutting down counts as permanent so the budget is not burned on expected restart noise.
- On the update path the synchronization stops at the last change it genuinely applied, so no content is skipped.
- Retrying is reported as a warning while it happens, and only becomes an error once the attempts run out, which is what the indexing side has always done.
- A retry is rebuilt from the operations captured *before* the request was submitted. The transport layer clears the entries of a request it has taken over, so re-sending that instance would re-send a list of nulls.
- The separate, narrower notion of a retryable failure that only the single-document path used is now removed.

**`Constants.java`**: five message constants for the new `WARN` lines, alongside the existing bulk ones.

## Results and Evidence

| # | Verified | How | Result |
|---|---|---|---|
| 1 | Unit test works as expected | `./gradlew test` | 12/12 |
| 2 | Plugin build | `./gradlew build` | `BUILD SUCCESSFUL in 4m 2s` |
| 3 | The nightly's failure reproduces | unpatched jar, live AIO, shed window | aborts **280 ms** in, 3 `ERROR` lines, zero retries |
| 4 | A shed write is retried and the sync survives | patched, same probe, block lifted on the first retry | 1 `WARN`, **0** `ERROR`, consumer reaches its offset |
| 5 | Only the shed items are re-sent | unit | retry batch = 1 of 2 items |
| 6 | Backoff matches the index path | live, exhausted budget | 1.016 s → 2.045 s → 4.015 s, **7.076 s** total |
| 7 | A permanent failure still aborts at once | live, 403 block | single attempt, **0** retry lines, 375 ms |
| 8 | Retry logs at `WARN`, not `ERROR` | capturing appender in unit test | 1 `WARN`, 0 `ERROR` |
| 10 | A real topology failure is classified as such | live AIO, snapshot path, shadow index dropped mid-load | `deferred 999 document(s) after a cluster-topology change; retry 1/5 in 5000ms`, and **0** `under load` lines |
| 11 | The topology budget is 5 retries over 95 s | unit, against a real 503-reporting `UnavailableShardsException` | 6 attempts (1st one + 5 retries), delays `5 s, 10 s, 20 s, 30 s, 30 s` |
| 12 | Nothing is committed over an incomplete load | live AIO, snapshot discarded | `local_offset` not advanced, live index untouched, doc counts identical before and after |

## Artifacts Affected

Content Manager Plugin

## Configuration Changes

NA

## Documentation Updates

NA

## Tests Introduced

Twelve unit tests in `ContentIndexTests`:

| Test | What it pins |
|------|--------------|
| `testBatchUpdate_RetriesShedBulkItemsAndSucceeds` | Only the shed item is re-submitted (not the whole batch), the consumer reaches its offset, and exactly one `WARN` and zero `ERROR` are logged |
| `testBatchUpdate_AbortsOnPermanentBulkFailure` | A rejected document still aborts the batch — the offset must not be committed past it |
| `testBatchUpdate_AbortsWhenClusterKeepsSheddingTheBulk` | The budget is finite: 1 initial attempt + 3 retries, then throw rather than drop |
| `testBatchUpdate_RetriesShedMultiGet` | The read that opens `batchUpdate` is retried, not just the write |
| `testBatchUpdate_RetriesWhenWholeBulkRequestIsRejected` | A request-level rejection (`ExecutionException`) is the same condition as a shed item |
| `testExecuteBulk_RetriesIndexNotFoundInsteadOfDroppingTheSnapshot` | The reported case: an index recreated mid-load is re-submitted, nothing is dropped |
| `testExecuteBulk_TopologyBudgetOutlastsARollingRestart` | 6 attempts with delays `5 s, 10 s, 20 s, 30 s, 30 s` — the schedule is asserted, not just the count |
| `testExecuteBulk_MixedBatchUsesTheTopologySchedule` | A batch holding both kinds of failure takes the longer schedule |
| `testExecuteBulk_DoesNotRetryNodeClosed` | A node shutting down stays permanent, so the budget is not burned on restart noise |
| `testExecuteBulk_DoesNotRetryForbiddenClusterBlock` | A 403 block stays permanent, guarding the decision to classify blocks by status rather than type |
| `testBatchUpdate_RetriesTopologyFailureOnTheLongerBudget` | The update path routes topology failures through the same schedule |
| `testExecuteBulk_RetrySurvivesTheTransportClearingTheRequest` | A retry is rebuilt from operations captured before submission, never from the cleared request |

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


MauriFont (approved):
## Review

LGTM on the approach — the three-way policy disagreement is real and the fix addresses it at the right level. Ran `ContentIndexTests` on the branch (38 tests, 0 failures) and audited the validation pack's captures rather than re-running the rig; every figure quoted in the report checks out against the raw files. Four comments.

**The before/after pair proves more than the report claims for it.** Both cases surface as the *same* exception type — `ClusterBlockException` at 429 retries 1/2/4 s, at 403 aborts on the first attempt. That's a live demonstration that classifying by `RestStatus` rather than by exception type was correct, and that adding `ClusterBlockException` to the retryable list (as suggested on #1913) would have made an administrative write block retry forever. Worth keeping that contrast in the PR description — it justifies a decision that otherwise reads as arbitrary.

### 1. The fix does not fully close #1900 — follow-up issue, not a blocker

`UpdateServiceImpl.findIndexForId` (line 338) calls `ContentIndex.exists()` (line 433) per document, and that's a bare synchronous get with no retry wrapper and no classifier. A `CircuitBreakingException` from there aborts the CVE sync exactly as #1900 describes — same symptom, same failure family, through a door this PR leaves open. On a node with the heap at its limit it's as exposed as the bulk that now retries.

Agreed it's out of scope here, but I'd open the follow-up **before** closing #1900 and note in the description that the fix covers the bulk path, not the whole update flow. Otherwise the next occurrence of those three ERROR lines looks like a regression of a closed issue.

### 2. The `IndexNotFoundException` trade-off is stated against a stale number

It was assessed on #1900 as "three retries cost roughly seven seconds" — that was the SHED budget. `retryPolicyFor` now routes it to `TOPOLOGY`, so a genuinely missing index costs **95 s**, and `mostPersistent` applies the same to any mixed batch. The conclusion may still hold; please restate it against the real figure so reviewers are agreeing to 95 s rather than 7 s.

### 4. Question: should the budgets be configurable?

`RetryPolicy` is a private enum with hardcoded constants. Not asking for settings on principle, but an operator on a cluster that needs longer than 95 s to settle has no lever short of a rebuild, and neither budget can be disabled to isolate a problem. Considered and rejected, or just not reached for?

Issue #1584, opened 2026-09-18
Use the CTI API setting as the base url for the CTI console


MauriFont:
**Is your feature request related to a problem? Please describe.**

The CTI Console base URL is hardcoded in the Content Manager plugin and cannot be overridden from `opensearch.yml`.

In `wazuh-indexer-plugins/plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/console/client/ApiClient.java:52`:

```java
private static final String BASE_URI = "https://api.pre.cloud.wazuh.com";
private static final String API_PREFIX = "/api/v1";
private static final String TOKEN_URI = BASE_URI + API_PREFIX + "/instances/token";
private static final String PRODUCTS_URI = BASE_URI + API_PREFIX + "/instances/me";
private static final String RESOURCE_URI = BASE_URI + API_PREFIX + "/platform/environments/token/exchange";
private static final String ENVIRONMENTS_ME_URI = BASE_URI + API_PREFIX + "/platform/environments/me";
```

The value is a pre-production host baked into the artifact. Every deployment — including a release build — performs instance registration, product/environment lookup and resource-token exchange against `api.pre.cloud.wazuh.com`, and the only way to point it somewhere else (production, a staging environment, an on-premises or proxied CTI endpoint) is to recompile the plugin.

These constants are `static final`, so they are resolved at class-load time, before any `Settings` object exists. That is the root of the problem: nothing that reads configuration can intervene.

The rest of the CTI client is already configurable through `plugins.content_manager.cti.api` (`PluginSettings.java:112-119`), which backs `getCtiBaseUrl()`. Two methods below the hardcoded constants, in the very same class, `getCatalogPlans()` already does the right thing (`ApiClient.java:310`, `:334`):

```java
String url = PluginSettings.getInstance().getCtiBaseUrl() + CATALOG_PLANS_PATH;
```

The same setting also backs the catalog client (`cti/catalog/client/ApiClient.java:70`, `:225`) and the telemetry ping (`cti/console/client/TelemetryClient.java:45`). The Console endpoints are the only CTI requests in the plugin that ignore it.

[...]

PR #1587, opened 2026-09-18
Change hardcoded operational constants to plugin settings

### LINKED ISSUE #1585: Change hardcoded constants to plugin settings
## Description

Both plugins already have a settings infrastructure: `PluginSettings` + `Plugin.getSettings()`, with `plugins.content_manager.*` and `plugins.setup.*` keys, bounded ranges and `NodeScope`/`Dynamic` properties.

A number of values that govern runtime behavior never made it into that infrastructure and are compiled-in constants, retry budgets, back off schedules, wait timeouts, batch and page sizes.
They cannot be changed without rebuilding the plugin

For example, `ContentIndex.RetryPolicy` bakes the whole bulk retry schedule into the enum
constructor:

```java
// plugins/content-manager/.../cti/catalog/index/ContentIndex.java
SHED(3, 1_000, 30_000, ...),
TOPOLOGY(5, 5_000, 30_000, ...);
```

The task is to go through the code, find the constants that are really operational configuration, and add settings for them

## Functional requirements

- Study the compiled-in constants in both plugins and identify which ones are operational
  configuration an operator may need to tune, versus which ones are legitimately constants
- Expose the former as settings following the existing conventions, with a sensible `min`/`max`
  range each
- Defaults must equal today's constants: an untouched `opensearch.yml` behaves exactly as the
  current release does

- Follow the existing conventions: `Setting` declared in `PluginSettings`, registered in
  `getSettings()`, `DEFAULT_*` constant ...
- No behavior change at default values

## Plan

- [ ] Study the constants across both plugins and triage them into promote to setting / keep as constant, recording the rationale in the PR.
- [ ] Add the settings for the promoted values.
- [ ] Unit tests: default, range clamping, and dynamic re-read where applicable.
- [ ] Update the reference documentation settings tables.

Jorgesnchz:
## Description

Both plugins already have a settings infrastructure, `PluginSettings` + `Plugin.getSettings()`, with `plugins.content_manager.*` / `plugins.setup.*` keys, bounded ranges and `NodeScope`/`Dynamic` properties. A number of values that govern runtime behaviour never made it into it and were still compiled-in constants: retry budgets, backoff schedules, wait timeouts, batch and page sizes, and one remote endpoint. They could not be changed without rebuilding the plugin.

This PR creates 26 new settings. Each one defaults to exactly the constant it replaced, so an untouched `opensearch.yml` behaves as the current release does. 

Closes https://github.com/wazuh/wazuh-indexer-plugins/issues/1585

Now user's environment are not obliged to match the values we have defined.
Example: An installation on slow or undersized hardware can now solve its own problem by editing `opensearch.yml` or calling the cluster settings API

> [!NOTE]
>  This PR will have conflics with [#1584](https://github.com/wazuh/wazuh-indexer-plugins/issues/1584), sonce both issues work on `cti/console/client/ApiClient.java` file, we will need to adapt the one that is merged later to the latest changes

## Proposed Changes

- Add the new settings and modify the code to include them
- Modify the documentation related to the settings and include the new settings created

#### Unit tests

```
content-manager: 780 tests, 0 failures, 0 skipped
setup:           100 tests, 0 failures, 0 skipped
```

```
BUILD SUCCESSFUL in 1m 15s
15 actionable tasks: 3 executed, 12 up-to-date
```

#### Live validation

Validation error for `plugins.content_manager.search_page_size: 99999` in `opensearch.yml`:

Visibility did not widen scope, the static settings are still rejected dynamically:

``` console
=== 3. static settings are rejected by PUT _cluster/settings
PASS  plugins.content_manager.search_page_size rejected 400, message names the key
      persistent setting [plugins.content_manager.search_page_size], not dynamically updateable
PASS  plugins.content_manager.cti.console.api rejected 400, message names the key
      persistent setting [plugins.content_manager.cti.console.api], not dynamically updateable
...
=== totals: 14 pass, 0 fail
```

Dynamic updates and range boundaries:

``` console
=== 2. dynamic settings accept PUT _cluster/settings and read the value back
=== totals: 12 pass, 0 fail, 0 missing-from-API

=== 4. range sweep: min / max / min-1 / max+1 on every bounded dynamic setting
=== totals: 48 pass, 0 fail, 0 missing-from-API
```

### Artifacts Affected

- Content Manager plugin
- Setup plugin

### Configuration Changes

26 new settings, all optional, all defaulting to the previously compiled-in value

### Documentation Updates

- `docs/ref/modules/content-manager/configuration.md`
- `docs/ref/modules/setup/configuration.md` 

### Tests Introduced

- Content Manager: `PluginSettingsTests` 
- Setup: `PluginSettingsTests`

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/SpaceService.java
```diff
@@ -181,7 +181,7 @@ private void collectDeleteRequestsAsync(
                                     // match every document in the index would delete it, and silently lose
                                     // the user's policy settings and the filters they created.
                                     sourceBuilder.query(QueryBuilders.termQuery(Constants.Q_SPACE_NAME, spaceName));
-                                    sourceBuilder.size(10000);
+                                    sourceBuilder.size(PluginSettings.getInstance().getSearchPageSize());
```
ferooo23: The `10000` this replaces was not a page size: it is the `index.max_result_window` default, i.e. "fetch everything". This search has no `search_after` or PIT, so a lowered `search_page_size` truncates it silently, deleting a space would leave every document past the cap behind, still carrying the `space.name` of a space that no longer exists, and the request still returns 200.

Same shape at L404, L475, L531, L1100, and in `DetectorLookupService` L122 and L216. The two `ConsumerIocService` sites are the only real pagination (PIT + `search_after`), so I would keep the setting there and restore a named constant here:

```suggestion
                                    sourceBuilder.size(MAX_RESULT_WINDOW);
```

plugins/content-manager/src/main/java/com/wazuh/contentmanager/engine/client/EngineSocketClient.java
```diff
@@ -50,16 +51,14 @@
...
-            "/usr/share/wazuh-indexer/engine/sockets/engine-api-http.sock";
     private static final int BUFFER_SIZE = 8192;
     private static final ObjectMapper objectMapper = new ObjectMapper();
 
     private final String socketPath;
 
     /** Creates a EngineSocketClient with the default socket path. */
     public EngineSocketClient() {
-        this(DEFAULT_SOCKET_PATH);
+        this(PluginSettings.getInstance().getEngineSocketPath());
```
ferooo23: Commenting here because `EngineSocketClientTests` is not in the diff, but that is where the fix goes.

This line makes the no-arg constructor depend on the settings singleton, and that test never initializes it, so `testDefaultConstructorUsesDefaultGetSocketPath` now fails with `IllegalStateException: Plugin settings have not been initialized.` I get 780 tests, 1 failed on this branch, same count as the PR description, different result; it is order-dependent, so it can pass depending on which class ran before it in the same JVM. `PluginSettings.getInstance(Settings.EMPTY)` in `setUp` plus an `@After` reset fixes it. Production is unaffected, the singleton is initialized in `createComponents` before anything uses it.

Jorgesnchz:
I've seen that the tests are broken, on my environment they work as expected, so I believe that the error is related to the workflow or its configuration, I'll take a deeper look as soon as I have the chance to

``` console
jorge@jorge-wazuh:~/wazuh/wazuh-indexer-plugins/plugins/content-manager$ ./gradlew build
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 9.5.1
...
> Task :yamlRestTest NO-SOURCE
> Task :check
> Task :build

BUILD SUCCESSFUL in 4m
53 actionable tasks: 39 executed, 14 up-to-date
17:41:11: Execution finished 'build'.
jorge@jorge-wazuh:~/wazuh/wazuh-indexer-plugins/plugins/content-manager$ ./gradlew check
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 9.5.1
...
> Task :yamlRestTest NO-SOURCE
> Task :check
BUILD SUCCESSFUL in 3m 43s
49 actionable tasks: 9 executed, 1 from cache, 39 up-to-date
17:47:21: Execution finished 'check'. # 1080/108 passed
jorge@jorge-wazuh:~/wazuh/wazuh-indexer-plugins/plugins/content-manager$ git branch --show-current
enhancement/1585-move-to-settings-hardcoded-constants

```

plugins/content-manager/src/main/java/com/wazuh/contentmanager/settings/PluginSettings.java
```diff
@@ -119,6 +170,15 @@ public class PluginSettings {
                     Setting.Property.NodeScope,
                     Setting.Property.Filtered);
 
+    /** Request timeout, in seconds, for CTI API calls. */
+    public static final Setting<Integer> CTI_API_TIMEOUT =
```
ferooo23, line 174: Three things on this one.

It is not in `ContentManagerPlugin#getSettings()`, I count 52 settings declared and 51 registered, and this is the one missing. An unregistered `plugins.*` key in `opensearch.yml` stops the node from starting, so anyone who tries to use it gets a node that will not boot.

The default also changes behaviour. The Console client had `private final int TIMEOUT = 5` both before and after #1589, and it now reads `getCtiRequestTimeout()`, which defaults to 60, twelve times longer on an untouched `opensearch.yml`. `PluginSettingsTests` L485 still asserts `5`, so the test encodes the intent and the default drifted from it.

One more on the naming: after this, `client.timeout` (10s) is the catalog client and `cti.api.timeout` (60s) is the Console client. The names point at each other's meaning, and an operator raising "the CTI API timeout" would tune the wrong one.

plugins/content-manager/src/test/java/com/wazuh/contentmanager/settings/PluginSettingsTests.java
```diff
@@ -449,4 +450,195 @@ public void testSetAndGetCatalogSyncInterval() {
...
+        // Constants.MAX_LOCK_ACQUIRE_RETRIES / LOCK_ACQUIRE_RETRY_BACKOFF_MILLIS /
+        // LOCK_STALE_THRESHOLD_MILLIS
+        Assert.assertEquals(20, s.getResourceLockMaxRetries());
+        Assert.assertEquals(100L, s.getResourceLockRetryBackoffMillis());
+        Assert.assertEquals(30_000L, s.getResourceLockStaleThresholdMillis());
+        // Constants.MAX_USER_OVERRIDES_UPDATE_ATTEMPTS / IntegrationService.MAX_RETRIES
+        Assert.assertEquals(3, s.getUserOverridesMaxUpdateAttempts());
+        Assert.assertEquals(5, s.getIntegrationMaxUpdateAttempts());
+        // cti.console.client.ApiClient BASE_URI / TIMEOUT
+        Assert.assertEquals("https://api.pre.cloud.wazuh.com", s.getCtiBaseUrl());
```
ferooo23, line 484: `getCtiBaseUrl()` returns `CTI_API_URL`, whose default is `https://api.pre.cloud.wazuh.com/api/v1` , the `/api/v1` is part of the setting, which is what makes appending the bare paths work in `ApiClient`. So this expects the base URL without the prefix and gets it with:

```
expected:<...pre.cloud.wazuh.com[]> but was:<...pre.cloud.wazuh.com[/api/v1]>
```

`testPromotedSettingsReadCustomValues` has the same problem plus one of its own: L528-L529 still configure `plugins.content_manager.cti.console.api` and `plugins.content_manager.cti.console.timeout`, which no longer exist, so nothing is applied and the assertions run against the defaults.

Worth noting L485 (`assertEquals(5, s.getCtiRequestTimeout())`) never runs, because L484 fails first , it is a fourth failure hiding behind this one, and the one that disagrees with the new `60` default.

plugins/content-manager/src/main/java/com/wazuh/contentmanager/engine/client/EngineSocketClient.java
ferooo23, line 61: Commenting here because `EngineSocketClientTests` is not in the diff, but that is where the fix goes.

This line makes the no-arg constructor depend on the settings singleton, and that test never initializes it, so `testDefaultConstructorUsesDefaultGetSocketPath` fails with `IllegalStateException: Plugin settings have not been initialized.` It is order-dependent, so it can pass depending on which class ran before it in the same JVM ,it is one of the three failures I get on this branch. `PluginSettings.getInstance(Settings.EMPTY)` in `setUp` plus an `@After` reset fixes it. Production is unaffected, the singleton is initialized in `createComponents` before anything uses it.

docs/ref/modules/content-manager/configuration.md
```diff
@@ -31,6 +31,32 @@ The Content Manager plugin is configured through settings in `opensearch.yml`. A
 - **`plugins.content_manager.max_kvdbs`** (Integer, default `100`, minimum `0`, no upper bound, dynamic) — maximum number of KVDBs that can be created. Requests that would exceed this limit are rejected with HTTP 400.
 - **`plugins.content_manager.max_filters`** (Integer, default `100`, minimum `0`, no upper bound, dynamic) — maximum number of filters that can be created per space. Requests that would exceed this limit are rejected with HTTP 400.
 
+- **`plugins.content_manager.cti.console.api`** (String, default `https://api.pre.cloud.wazuh.com`) — base URL for the Wazuh CTI Console, used for instance registration and token exchange. Distinct from `plugins.content_manager.cti.api`, which addresses the CTI catalog API.
```
ferooo23, line 34: These two settings no longer exist after the merge, so an operator copying them into `opensearch.yml` gets a node that will not start , the same failure mode as the unregistered `cti.api.timeout`, but reached by following the documentation. `plugins.content_manager.cti.api.timeout` is not documented anywhere either, so the list is out in both directions.

plugins/content-manager/src/main/java/com/wazuh/contentmanager/cti/catalog/service/SpaceService.java
ferooo23, line 184: The `10000` this replaces was not a page size: it is the `index.max_result_window` default, i.e. "fetch everything". This search has no `search_after` or PIT, so a lowered `search_page_size` truncates it silently , deleting a space would leave every document past the cap behind, still carrying the `space.name` of a space that no longer exists, and the request still returns 200.

Same shape at L404, L475, L531, L1100, and in `DetectorLookupService` L122 and L216. The two `ConsumerIocService` sites are the only real pagination (PIT + `search_after`), so I would keep the setting there and restore a named constant here:

```suggestion
                                    sourceBuilder.size(MAX_RESULT_WINDOW);
```

plugins/setup/src/main/java/com/wazuh/setup/index/IndexStateManagement.java
```diff
@@ -120,12 +120,13 @@ void indexPolicy(String policy) {
                 Exception
                         e) { // TimeoutException may be raised by actionGet(), but we cannot catch that one.
             // Exit condition. Re-attempt to create the index also failed. Original exception is rethrown.
-            if (!this.retry_index_creation) {
+            if (this.indexCreationAttempts
```
ferooo23, line 123: `indexCreationAttempts` is a single instance field, and `createPolicies()` is `this.policies.forEach(this::indexPolicy)` over six policies with no reset between them, so this budget is shared rather than per task.

With `max_retries: 3`: if `events` spends two re-attempts and `findings` spends one, then `raw_events` throws on its very first timeout, and `active_responses`, `metrics` and `ai_assistant_sessions` never run at all. The operator raised the number and five of the six policies got zero retries.

This matches the old `retry_index_creation` boolean exactly, so defaults are preserved , the issue is only that the setting now promises something it does not deliver. Resetting the counter at the top of `indexPolicy` would make the behaviour match the doc entry, which reads as per task.

plugins/content-manager/src/main/java/com/wazuh/contentmanager/settings/PluginSettings.java
```diff
@@ -433,7 +493,288 @@ public class PluginSettings {
...
+
+    /**
+     * Age, in milliseconds, past which a held resource-creation lock is treated as orphaned by a
+     * crashed node and stolen by the next caller.
+     */
+    public static final Setting<Long> RESOURCE_LOCK_STALE_THRESHOLD_MILLIS =
+            Setting.longSetting(
+                    "plugins.content_manager.resource_lock.stale_threshold_millis",
+                    DEFAULT_RESOURCE_LOCK_STALE_THRESHOLD_MILLIS,
+                    1_000,
```
ferooo23, line 630: This one decides whether a lock held by another node is dead, and it is `NodeScope`, so it comes from each node's own `opensearch.yml`. It used to be a compiled constant, identical everywhere; now node A can hold the lock believing it has 60s left while node B reads the same `acquired_at` against its own 30s and steals it. Both are then inside the critical section the mutex exists to protect, and `release` deletes by id with no ownership check, so A's release also drops B's lock.

The 1s minimum has a second problem on its own: the holder never renews `acquired_at`, so the threshold has to outlast the whole resource-creation chain. Below that, every concurrent request steals the lock and the mutex stops working: uniformly, no disagreement needed.

```suggestion
                    5_000,
```

Could the entry in `configuration.md` say that this value must be identical on every node and above the worst-case resource creation time?

plugins/content-manager/src/main/java/com/wazuh/contentmanager/settings/PluginSettings.java
```diff
@@ -433,7 +493,288 @@ public class PluginSettings {
...
+
+    /**
+     * How long, in seconds, to wait for the deletion of stale Security Analytics rules and
+     * integrations after a content swap.
+     */
+    public static final Setting<Integer> SA_CLEANUP_TIMEOUT_SECONDS =
+            Setting.intSetting(
+                    "plugins.content_manager.security_analytics.cleanup_timeout_seconds",
+                    DEFAULT_SA_CLEANUP_TIMEOUT_SECONDS,
+                    1,
```
ferooo23, line 722: `deleteStaleResources` is the only place in `ConsumerRulesetService` where the `await()` result is discarded , L763 and L785, against L435, L566 and L677 which all check it and flip `success`. On timeout the method carries on, logs `Deleted {} stale rule(s)` with the size of the set it *intended* to delete, and the `finally` clears `preSwapRuleIds`, so what was left undeleted is never retried and stays in Security Analytics.

That is all pre-existing, and I would not ask you to fix it here. But at 120s fixed it took a real Security Analytics stall to reach; a 1s minimum makes it reachable by configuration. Raising the floor to the default keeps it exactly as reachable as it is today, and the repo already does this in `client.timeout` (10, range 10–50), `max_concurrent_bulks` (5, range 1–5) and `max_items_per_bulk` (1000, range 10–1000):

```suggestion
                    120,
```

plugins/content-manager/src/main/java/com/wazuh/contentmanager/settings/PluginSettings.java
```diff
@@ -433,7 +493,288 @@ public class PluginSettings {
...
+                    100,
+                    600_000,
+                    Setting.Property.NodeScope,
+                    Setting.Property.Dynamic);
+
+    /**
+     * Ceiling, in milliseconds, for the exponential backoff between topology-deferred bulk
+     * re-submissions.
+     */
+    public static final Setting<Long> BULK_TOPOLOGY_MAX_BACKOFF_MILLIS =
```
ferooo23, line 566: The three bulk-retry bounds are independent, and their worst case is the product: `max_retries` × `max_backoff_millis`, because the initial delay is already clamped to the ceiling. At the maxima this PR allows that is 20 × 600 000 ms = **200 minutes**, against the 95s that is fixed today (5s + 10s + 20s + 30s + 30s).

That matters because two of the three retry paths block: `retryOrDrop` schedules on the thread pool and is fine at any value, but `executeBulkUpdate` L750 and `retryTransient` L1056 call `Thread.sleep`, and the sync pass runs on the shared pool (`CatalogSyncJob` L132, `threadPool.generic().execute(...)`).

Nothing in the enum's own rationale supports ten minutes between retries — TOPOLOGY says "a rolling restart takes tens of seconds per node", SHED says "pressure clears in seconds". Lowering `max_retries` to 10 and both backoff maxima to 60 000 brings the worst case to 10 minutes, still 6× the headroom over today. Ranges can be widened later but only narrowed before release, so this is the free moment:

```suggestion
                    60_000,
```

Same change on `bulk.retry.shed.max_backoff_millis`, the two `initial_backoff_millis` (no point allowing an initial above the ceiling maximum, the code clamps it), and both `max_retries`.

docs/ref/modules/content-manager/configuration.md
```diff
@@ -31,6 +31,32 @@ The Content Manager plugin is configured through settings in `opensearch.yml`. A
...
+- **`plugins.content_manager.security_analytics.detector_interval`** (Integer, default `2`, range 1–10080, dynamic) — detector schedule interval, in minutes, applied when the CTI integration document does not specify one, or specifies one outside the bounds Ruleset Management accepts.
+- **`plugins.content_manager.update_sub_batch_size`** (Integer, default `50`, range 1–1000) — maximum number of UPDATE offsets batched into a single MultiGet + bulk request while applying a catalog changeset.
+- **`plugins.content_manager.offset_flush_interval`** (Integer, default `10`, range 1–1000) — how many batches (changeset application) or bulks (snapshot load) are processed between consumer-offset checkpoints. A smaller value narrows the window of work replayed after a crash, at the cost of more writes to the consumer-state document.
+- **`plugins.content_manager.search_page_size`** (Integer, default `10000`, range 100–10000) — page size used by the internal paginated searches that enumerate content documents (space membership, IoC reconciliation, detector lookup).
```
ferooo23, line 58: Seven of the nine call sites have no pagination, so this describes the setting as safer to lower than it is.

```suggestion
- **`plugins.content_manager.search_page_size`** (Integer, default `10000`, range 100–10000) — page size used by the paginated IoC reconciliation scan. Lower values mean more round trips over the same documents, not fewer documents.
```

docs/ref/modules/setup/configuration.md
```diff
@@ -4,4 +4,5 @@ The Setup plugin is configured through settings in `opensearch.yml`. All setting
 
 - **`plugins.setup.timeout`** (Integer, default `30`) — timeout in seconds for index and search operations.
 - **`plugins.setup.backoff`** (Integer, default `15`) — delay in seconds for the retry mechanism involving initialization tasks.
+- **`plugins.setup.max_retries`** (Integer, default `1`, range 0–10) — number of times an initialization task (index or index-template creation) that timed out is re-attempted, each attempt separated by `plugins.setup.backoff`, before the failure is rethrown and the node shuts down. `0` disables retrying. Raise it on clusters where a cluster-manager election, and therefore the first index creation, routinely outlasts the default single re-attempt.
```
ferooo23, line 7: The budget is shared across the ISM index and its six policies rather than per task, and nothing in the plugin shuts the node down , `SetupPlugin` L276-L282 catches the exception, marks setup `failed` and skips every index registered after the ISM one.

```suggestion
- **`plugins.setup.max_retries`** (Integer, default `1`, range 0–10) — number of times an initialization task that timed out is re-attempted, each attempt separated by `plugins.setup.backoff`. `0` disables retrying. When the budget runs out the exception is rethrown: initialization stops, every index after the failing one is skipped, and the setup status is marked `failed`. Raise it on clusters where a cluster-manager election, and therefore the first index creation, routinely outlasts the default single re-attempt.
```

docs/ref/modules/setup/configuration.md
```diff
@@ -4,4 +4,5 @@ The Setup plugin is configured through settings in `opensearch.yml`. All setting
 
 - **`plugins.setup.timeout`** (Integer, default `30`) — timeout in seconds for index and search operations.
 - **`plugins.setup.backoff`** (Integer, default `15`) — delay in seconds for the retry mechanism involving initialization tasks.
```
ferooo23, line 6: Both are bounded in code (5–120 and 5–60) but the ranges are not documented, so `plugins.setup.timeout: 3` gives a node that refuses to start with nothing here to warn about it , the new neighbour below does carry its range.

```suggestion
- **`plugins.setup.timeout`** (Integer, default `30`, range 5–120) — timeout in seconds for index and search operations.
- **`plugins.setup.backoff`** (Integer, default `15`, range 5–60) — delay in seconds for the retry mechanism involving initialization tasks.
```

ferooo23 (changes requested):
The settings work itself is solid: every `Dynamic` setting has an update consumer and a `volatile` field, the documented values match the code.

A few things, though. On the branch as it stands I get **782 tests, 3 failed**. Comments inline.

**`PluginSettings.java`**

- `CTI_API_TIMEOUT` is declared but not registered in `getSettings()`, and its default moves the Console request timeout from 5s to 60s.
- The lock staleness threshold is `NodeScope` on a cluster-wide mutex, so two nodes can disagree about whether a held lock is alive.
- The Security Analytics cleanup timeout bottoms out at 1s.
- The three bulk-retry bounds combine into 200 minutes of a blocked GENERIC thread.

**`PluginSettingsTests.java`**

- Two assertions were repointed at `getCtiBaseUrl()`, which returns the URL with `/api/v1`, and the custom-value test still configures the two removed `cti.console.*` keys.

**`EngineSocketClient.java`**

- The no-arg constructor needs the settings singleton, and `EngineSocketClientTests` does not initialize it.

**`SpaceService.java`**

- `search_page_size` caps seven searches that have no pagination, so lowering it silently truncates them.

**`IndexStateManagement.java`**

- The retry budget is shared by the ISM index and its six policies rather than being per task.

**Docs**

- `content-manager/configuration.md` — two removed settings are still listed, `cti.api.timeout` is missing, the `search_page_size` description, and no `PUT _cluster/settings` example for the new dynamic keys.
- `setup/configuration.md` — no ranges on `timeout` and `backoff`, and the `max_retries` description.
- `content-manager/architecture.md` and `content-manager/troubleshooting.md` — three values that are now settings are still stated as fixed. Neither file is in the diff, so I could not leave those inline.

ferooo23:
## Validation 🟢

Single-node AIO (5.0.0, OpenSearch 3.6.0, content `rc1-t1-ruleset-5`), with `wazuh-indexer-setup` and `wazuh-indexer-content-manager` built from this branch plus the ISM fix. Three of the new static settings are set in `opensearch.yml` to non-default values:

```yaml
plugins.content_manager.cti.console.timeout: 10
plugins.setup.max_retries: 2
plugins.content_manager.search_page_size: 5000
```

### Settings

The node starts with them (an unregistered key would stop it) and applies them. The rest keep the constants they replace, here `RetryPolicy.TOPOLOGY(5, 5_000, 30_000)`:

```
GET _cluster/settings?include_defaults=true&filter_path=defaults.plugins.setup.max_retries,defaults.plugins.content_manager.cti.console.timeout,defaults.plugins.content_manager.search_page_size,defaults.plugins.content_manager.bulk.retry.topology
```
```json
{
  "defaults": {
    "plugins": {
      "content_manager": {
        "cti": { "console": { "timeout": "10" } },
        "bulk": {
          "retry": {
            "topology": {
              "max_retries": "5",
              "initial_backoff_millis": "5000",
              "max_backoff_millis": "30000"
            }
          }
        },
        "search_page_size": "5000"
      },
      "setup": { "max_retries": "2" }
    }
  }
}
```

All 25 settings added by this PR are present on the node. Bounds and scope are enforced by the cluster settings API (keys under `plugins.content_manager.`):

| `PUT _cluster/settings` (`persistent`) | Response |
|---|---|
| `bulk.retry.topology.max_retries: 8` | `200`, read back as `8` |
| `bulk.retry.topology.max_retries: 11` | `400` must be <= 10 |
| `bulk.retry.topology.max_backoff_millis: 60001` | `400` must be <= 60000 |
| `resource_lock.stale_threshold_millis: 4999` | `400` must be >= 5000 |
| `search_page_size: 500` | `400` not dynamically updateable |
| `security_analytics.cleanup_timeout_seconds: 200` | `400` not dynamically updateable |
| `cti.console.api`, `cti.api.timeout` | `400` not recognized |

### Setup

```
GET .wazuh-setup-status/_search?filter_path=hits.hits._source.status
GET _plugins/_ism/policies?filter_path=total_policies
```
```json
{"hits":{"hits":[{"_source":{"status":"ready"}}]}}
{"total_policies":6}
```

The six policies are created in ~160 ms with no `Retrying...` in the log. The retry path is covered by `testIndexPolicy_StopsAfterMaxRetries`, which fails against `73285443` ("still retrying after 20 attempts") and passes with the fix.

### Synchronization

A content update applies a real changeset with the new build:

```
POST _plugins/_content_manager/update
GET .wazuh-cti-consumers/_doc/cti:catalog:consumer:ruleset?filter_path=_source.status,_source.local_offset,_source.remote_offset,_source.pending_sync_phases
```
```json
{
  "_source": {
    "status": "ready",
    "local_offset": 1735,
    "remote_offset": 1735,
    "pending_sync_phases": []
  }
}
```

The ruleset moves from offset 1396 to 1735 and the other two consumers stay `ready`. Document counts in the seven `wazuh-threatintel-*` indices are identical before and after, since the 339 operations were updates.

ferooo23 (approved):
LGTM!

PR #1596, opened 2026-09-22
Resolve the plugins' build version from VERSION.json

### LINKED ISSUE #1595: Resolve the plugins' build version from VERSION.json
**Is your feature request related to a problem? Please describe.**

On the `5.0.0` branch, both plugins resolve the build version from a **hardcoded fallback** in their `build.gradle`:

- `plugins/setup/build.gradle:9` → `wazuh_version = System.getProperty("version", "5.0.0")`
- `plugins/content-manager/build.gradle:24` → `wazuh_version = System.getProperty("version", "5.0.0")`

This literal duplicates the value that already lives in `VERSION.json`, which creates two problems:

1. **Drift.** The fallback only applies to local builds (a bare `./gradlew` with no `-Dversion`; CI passes it explicitly), so a stale value is easy to miss. It is kept in sync only by `update_build_gradle_version()` in `tools/repository_bumper.sh`, i.e. by a side effect of the bumper rather than by the build itself. Live proof of the drift: on `main` today `VERSION.json` is `5.1.0` but `plugins/setup/build.gradle:9` still reads `"5.0.0"`.
2. **A recurring merge conflict on every upward merge.** The line is touched independently on each release branch, so `5.0.0 → 5.0.1 → main` conflicts on it every single time. It happened again in the current `5.0.1 → main` merge: `plugins/content-manager/build.gradle` conflicted between `main`'s dynamic read and `5.0.1`'s bumped literal, and the resolution is always "take the dynamic one" — pure noise.

`main` already solved this for one of the two plugins in `d2c6dc85` ("Dynamically read product version"), but `5.0.0` and `5.0.1` still carry the hardcoded literal, so the conflict keeps regenerating from the bottom of the chain.

**Describe the solution you'd like**

Backport the same auto-detection to `5.0.0`, in **both** `build.gradle` files, so the version has a single source of truth (`VERSION.json`) and the line stops diverging across branches.

Replace:

```groovy
    wazuh_version = System.getProperty("version", "5.0.0")
```

with the snippet already in use on `main` (`plugins/content-manager/build.gradle:24-25`):

```groovy
    wazuh_version = System.getProperty("version") ?: new groovy.json.JsonSlurper()
        .parse(new File(projectDir, "../../VERSION.json")).version
```

The relative path `../../VERSION.json` resolves correctly from both `plugins/setup/` and `plugins/content-manager/`. The `-Dversion` system property keeps taking precedence, so CI behaviour is unchanged; only the fallback path changes.

Files to touch on `5.0.0`:

-  `plugins/setup/build.gradle`
- `plugins/content-manager/build.gradle`
- `tools/repository_bumper.sh` — see additional context

Once merged, the change propagates naturally through `5.0.0 → 5.0.1 → main` and both branches converge on the same line, removing the conflict for good.

**Describe alternatives you've considered**

N/A

**Additional context**

`tools/repository_bumper.sh` needs attention as part of this change. Its `update_build_gradle_version()` function matches the literal with:

```bash
grep -qE 'System\.getProperty\("version", "[0-9]+\.[0-9]+\.[0-9]+"\)' "$file"
```

Once a file uses the dynamic read, that pattern no longer matches and the function silently logs `Warning: hardcoded 'version' fallback not found in <file>; skipping sync.` and moves on. This is already happening on `main` for `plugins/content-manager/build.gradle`. Since `VERSION.json` becomes the single source of truth, the function should be removed (or reduced to a guard that fails loudly if a hardcoded fallback reappears), and the comment block above it updated accordingly.

MauriFont:
## Description

Both plugins resolved their build version from a hardcoded literal (`System.getProperty("version", "5.0.0")`) that duplicates `VERSION.json`. It only applies to local builds without `-Dversion`, so it drifts unnoticed, and it conflicts on every upward merge because each release branch rewrites it to a different value.

`main` already fixed this for one plugin in `d2c6dc85`. This backports it to `5.0.0` for both, so the change propagates through `5.0.0 → 5.0.1 → main` and the branches converge.

Closes #1595

## Proposed Changes

**Both `build.gradle` files now read `VERSION.json`**, using the snippet already on `main`:

```groovy
    wazuh_version = System.getProperty("version") ?: new groovy.json.JsonSlurper()
        .parse(new File(projectDir, "../../VERSION.json")).version
```

`-Dversion` still takes precedence, so CI is unaffected.

**`tools/repository_bumper.sh` guards instead of syncing.** Its `update_build_gradle_version()` matched the literal to `sed` it; once a file reads `VERSION.json` the pattern stops matching and the function becomes a silent no-op (already the case on `main` for content-manager). With nothing left to sync, it is now `check_build_gradle_version()`, which exits 1 if a literal reappears — kept as a guard because this line is a repeat merge-conflict offender.

### Results and Evidence

```console
$ cd plugins/setup && ./gradlew -q properties | grep "^version:"
version: 5.0.0.0
$ cd plugins/content-manager && ./gradlew -q properties | grep "^version:"
version: 5.0.0.0
$ cd plugins/setup && ./gradlew -q properties -Dversion=9.9.9 | grep "^version:"
version: 9.9.9.0

$ bash -n tools/repository_bumper.sh && echo OK
OK
$ grep -rn "update_build_gradle_version" tools/ .github/ ; echo "exit=$?"
exit=1
```

### Artifacts Affected

Plugin zip names and versions are unchanged: `5_builderpackage_plugins.yml` passes an explicit `-Dversion`, which still wins.

What changes is every build that does *not* pass it — `5_testintegration_gradlecheck.yml` and local builds — which now derive the version from `VERSION.json` instead of a literal that can be stale. That is the fix.

Release tooling also changes: the bumper no longer rewrites `build.gradle`, and aborts if it finds a hardcoded fallback.

### Configuration Changes

None.

### Documentation Updates

None

### Tests Introduced

None

## Review Checklist

- [x] Code changes reviewed


Jorgesnchz (approved):
LGTM

